PR #10, opened 2016-04-15
Convenience peripheral functions implementation.

Kacper20:
I've written convenience functions to `Peripheral` as an extension.
Motivation is to provide to user simpler interface and not force him to use every time `discover` methods.
With this, user can create for example enums for every device he wants to use like that:
```
enum XXXCharacteristic: String, CharacteristicIdentifier {
    case ManufacturerName = "2A29"
    case GenericCommunication = "XXXX"

    var UUID: CBUUID {
        return CBUUID(string: self.rawValue)
    }
    var service: ServiceIdentifier {
        switch self {
        case .ManufacturerName:
            return XXXService.DeviceInformation
        case .GenericCommunication:
            return XXXService.GenericCommunication
        }
    }
}
enum XXXService: String, ServiceIdentifier {
    case DeviceInformation = "180A"
    case GenericCommunication = "XXX"

    var UUID: CBUUID {
        return CBUUID(string: self.rawValue)
    }
}
```
Later, thanks to that, simple action like making connection and reading value shortens from:
```
peripheral.connect()
    .flatMap { Observable.from($0.discoverServices([serviceId])) }
    .flatMap { Observable.from($0.discoverCharacteristics([characteristicId])}
    .flatMap { $0.readValue }
    .subscribeNext {
        let data = $0.value
    }
```
to:
```
peripheral.connect()
    .flatMap { $0.readValueFromCharacteristicWithId(XXXCharacteristic.ManufacturerName)
    .subscribeNext {
        let data = $0.value
    }
```
Also, if characteristic / descriptor / service is available in cache - we use it.

### DIFF
--- a/RxBluetoothKit/DeviceIdentifiers.swift
+++ b/RxBluetoothKit/DeviceIdentifiers.swift
@@ -35,3 +35,4 @@ protocol CharacteristicIdentifier {
     var UUID: CBUUID { get }
     var service: ServiceIdentifier { get }
 }
+
--- a/RxBluetoothKit/DeviceIdentifiers.swift
+++ b/RxBluetoothKit/DeviceIdentifiers.swift
@@ -1,0 +1,37 @@
+// The MIT License (MIT)
+//
+// Copyright (c) 2016 Polidea
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in all
+// copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+// SOFTWARE.
+
+import CoreBluetooth
+
+protocol ServiceIdentifier {
+    var UUID: CBUUID { get }
+}
+
+protocol DescriptorIdentifier {
+    var UUID: CBUUID { get }
+    var characteristic: CharacteristicIdentifier { get }
+}
+
+protocol CharacteristicIdentifier {
+    var UUID: CBUUID { get }
+    var service: ServiceIdentifier { get }
+}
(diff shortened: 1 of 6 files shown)


RxBluetoothKit/DeviceIdentifiers.swift
```diff
@@ -0,0 +1,37 @@
...
+}
+
+protocol DescriptorIdentifier {
+    var UUID: CBUUID { get }
+    var characteristic: CharacteristicIdentifier { get }
+}
+
+protocol CharacteristicIdentifier {
+    var UUID: CBUUID { get }
+    var service: ServiceIdentifier { get }
```
literator: Question: is `service` or `characteristic` always mandatory? Or is `UUID` enough sometime? I'm asking because maybe you can make e.g. `service` optional.

Kacper20: It's necessary for way better performance of implementation.

RxBluetoothKit/Peripheral+Convenience.swift
```diff
@@ -0,0 +1,138 @@
...
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+// SOFTWARE.
+
+import Foundation
+import CoreBluetooth
+import RxSwift
+
+extension Peripheral {
+
+    private func getService(id: ServiceIdentifier) -> Observable<Service> {
```
literator:
```swift
func getService(id: ServiceIdentifier)
```
->
```swift
func serviceWithIdentifier(identifier: ServiceIdentifier)
```
?

RxBluetoothKit/Peripheral+Convenience.swift
```diff
@@ -0,0 +1,138 @@
...
+            if let services = self.services,
+                let service = services.findElement({ $0.UUID == id.UUID  }) {
+                return Observable.just(service)
+            } else {
+                return Observable.from(self.discoverServices([id.UUID]))
+            }
+        }
+    }
+
+    func getCharacteristic(id: CharacteristicIdentifier) -> Observable<Characteristic> {
```
literator:
```swift
func getCharacteristic(id: CharacteristicIdentifier)
```
->
```swift
func characteristicWithIdentifier(identifier: CharacteristicIdentifier)
```
?

RxBluetoothKit/Peripheral+Convenience.swift
```diff
@@ -0,0 +1,138 @@
...
+                    }) {
+                        return Observable.just(characteristic)
+                    } else {
+                        return Observable.from(service.discoverCharacteristics([id.UUID]))
+                    }
+            }
+        }
+    }
+
+    func getDescriptor(id: DescriptorIdentifier) -> Observable<Descriptor> {
```
literator:
```swift
func descriptorWithIdentifier(identifier: DescriptorIdentifier)
```
?

RxBluetoothKit/Peripheral+Convenience.swift
```diff
@@ -0,0 +1,138 @@
...
+                        let descriptor = descriptors.findElement({ $0.UUID == id.UUID }) {
+                        return Observable.just(descriptor)
+                    } else {
+                        return Observable.from(characteristic.discoverDescriptors([id.UUID]))
+                    }
+            }
+        }
+    }
+
+    func monitorWriteForCharacteristicWithId(id: CharacteristicIdentifier) -> Observable<Characteristic> {
```
literator: `...withIdentifier` - I'd suggest using full name if you need it

literator:
Ok, so you use `id`/`Id` instead of `identifier`/`Identifier`. I personally like the latter more. Most swift style guides recommend not using short forms but long of words.

But it's up to you.

literator:
I got the idea and I like it! No comments to implementation :+1:.

Kacper20:
That's good idea - i'll change it right away. Big thanks

PR #20, opened 2016-04-27
Semantic change to `setNotificationAndMonitorChanges` and README.

Kacper20:
Changes we've been talking about - we've decided to disable notifications on dispose in `setNotificationAndMonitor` method. I'm not entirely sure about explaining this in README in current form - we could discuss it.

Thanks!

### DIFF
--- a/README.md
+++ b/README.md
@@ -173,7 +173,9 @@ characteristic.setNotificationAndMonitorUpdates()
 		let newValue = $0.value
 	}
 ```
-If you are not interested anymore in updates, just use this:
+If you are not interested anymore in updates, just use dispose this subscription - it will automatically set notification state on `false`.
+When you want to set notification state without monitoring updates, you should subscribe to: `characteristic.setNotifyValue(true)` method. Similarly: if you want to manually set notification state on `false` : subscribe to `characteristic.setNotifyValue(false)`
+
 ```swift
 characteristic.setNotifyValue(false)
 	.subscribe { characteristic in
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -173,7 +173,9 @@ characteristic.setNotificationAndMonitorUpdates()
 		let newValue = $0.value
 	}
 ```
-If you are not interested anymore in updates, just use this:
+If you are not interested anymore in updates, just use dispose this subscription - it will automatically set notification state on `false`.
```
literator: Maybe: "it will automatically set notify value to `false`"?

Source/Peripheral.swift
```diff
@@ -327,9 +327,15 @@ public class Peripheral {
...
-            return Observable.of(monitorValueUpdateForCharacteristic(characteristic),
-            setNotifyValue(true, forCharacteristic: characteristic).ignoreElements()
-                .subscribeOn(CurrentThreadScheduler.instance)).merge()
+        return Observable.create { observer in
+            let disposable = Observable.of(self.monitorValueUpdateForCharacteristic(characteristic),
+                self.setNotifyValue(true, forCharacteristic: characteristic).ignoreElements()
+                    .subscribeOn(CurrentThreadScheduler.instance)).merge().subscribe(observer)
+            return AnonymousDisposable {
+                disposable.dispose()
+                self.peripheral.setNotifyValue(false, forCharacteristic: characteristic.characteristic)
```
literator: > When you enable notifications for characteristic’s value, the peripheral calls the peripheral:didUpdateNotificationStateForCharacteristic:error: method of its delegate object to notify your app when the characteristic’s value changes.

I'm not sure whether it's called when `false` is set. However it seems not very important in this context.

Kacper20:
Closing for now - this change is not aligned with our designed decisions.

PR #22, opened 2016-05-04
Changed peripheral and central manager state handling

literator:


Kacper20:
Ok, so as we've discussed - we need to consider starting state.

Kacper20:
Merged already,

PR #27, opened 2016-05-23
Feature/state restoration

Kacper20:
Hi,

It's first look at implementation. Waiting for comments & improve suggestions. Later we'll document it.

### DIFF
--- a/Source/AdvertisementData.swift
+++ b/Source/AdvertisementData.swift
@@ -35,7 +35,7 @@ public struct AdvertisementData {
 
      - parameter advertisementData: Core Bluetooth's advertisement data
      */
-    public init(advertisementData: [String:AnyObject]) {
+    init(advertisementData: [String:AnyObject]) {
         self.advertisementData = advertisementData
     }
 
--- a/Source/AdvertisementData.swift
+++ b/Source/AdvertisementData.swift
@@ -35,7 +35,7 @@ public struct AdvertisementData {
 
      - parameter advertisementData: Core Bluetooth's advertisement data
      */
-    init(advertisementData: [String:AnyObject]) {
+    public init(advertisementData: [String:AnyObject]) {
         self.advertisementData = advertisementData
     }
 
--- a/Source/BluetoothManager.swift
+++ b/Source/BluetoothManager.swift
@@ -399,6 +399,6 @@ public class BluetoothManager {
     func listenOnRestoredState() -> Observable<RestoredState> {
         return centralManager
             .rx_willRestoreState
-            .map { RestoredState(restoredStateDictionary: $0)  }
+            .map { RestoredState(restoredStateDictionary: $0, bluetoothManager: self) }
     }
 }
--- a/Source/BluetoothManager.swift
+++ b/Source/BluetoothManager.swift
@@ -85,6 +85,9 @@ public class BluetoothManager {
 	 code.
 
 	 - parameter queue: Queue on which bluetooth callbacks are received. By default main thread is used.
+
+     - parameter options: An optional dictionary containing initialization options for a central manager.
+     For more info about it please refer to [`Central Manager initialization options`](https://developer.apple.com/library/ios/documentation/CoreBluetooth/Reference/CBCentralManager_Class/index.html)
 	 */
 	convenience public init(queue: dispatch_queue_t = dispatch_get_main_queue(),
 	                        options: [String : AnyObject]? = nil) {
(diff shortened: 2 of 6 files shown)


Source/AdvertisementData.swift
```diff
@@ -35,7 +35,7 @@ public struct AdvertisementData {
 
      - parameter advertisementData: Core Bluetooth's advertisement data
      */
-    public init(advertisementData: [String:AnyObject]) {
+    init(advertisementData: [String:AnyObject]) {
```
literator: Cool. It doesn't have to be public :)

Source/BluetoothManager.swift
```diff
@@ -86,7 +86,8 @@ public class BluetoothManager {
 
 	 - parameter queue: Queue on which bluetooth callbacks are received. By default main thread is used.
 	 */
-	convenience public init(queue: dispatch_queue_t = dispatch_get_main_queue()) {
+	convenience public init(queue: dispatch_queue_t = dispatch_get_main_queue(),
```
literator: This is seamless for current users 👍

Source/RestoredState.swift
```diff
@@ -0,0 +1,55 @@
...
+import CoreBluetooth
+
+/**
+ Convenience class which helps reading state of restored BluetoothManager
+ */
+public struct RestoredState {
+
+    public let restoredStateData: [String:AnyObject]
+
+    let bluetoothManager: BluetoothManager
```
literator: I'm thinking that it might be useful for user to have access to raw bluetooth manager, however that wouldn't be very pure...

Source/RestoredState.swift
```diff
@@ -0,0 +1,55 @@
...
+        self.restoredStateData = restoredStateDictionary
+        self.bluetoothManager = bluetoothManager
+    }
+
+    /**
+     Creates restored state information based on CoreBluetooth's dictionary
+
+     - parameter restoredState: Core Bluetooth's restored state data
+     */
+    var peripherals: [Peripheral] {
```
literator: `public private(set)`?

Source/RestoredState.swift
```diff
@@ -0,0 +1,55 @@
...
+        let objects = restoredStateData[CBCentralManagerRestoredStatePeripheralsKey] as? [AnyObject]
+        guard let arrayOfAnyObjects = objects else { return [] }
+        return arrayOfAnyObjects.flatMap { $0 as? CBPeripheral }
+            .map { RxCBPeripheral(peripheral: $0) }
+            .map { Peripheral(manager: bluetoothManager, peripheral: $0) }
+    }
+
+    var scanOptions: [String : AnyObject]? {
```
literator: `public private(set)`?

Source/RestoredState.swift
```diff
@@ -0,0 +1,55 @@
...
+        return restoredStateData[CBCentralManagerRestoredStatePeripheralsKey] as? [String : AnyObject]
+    }
+
+    var services: [Service] {
```
literator: `public private(set)`?

literator:
Looks good, but please take a look at some comments.

Source/RestoredState.swift
Kacper20: Hm, every time user gets access to `RestoredState` instance it by calling `listenOnRestoredState() ` on BluetoothManager. So i think in most common use case it'll not be a problem.

Source/RestoredState.swift
```diff
@@ -0,0 +1,62 @@
...
+
+/**
+ Convenience class which helps reading state of restored BluetoothManager
+ */
+public struct RestoredState {
+
+    /**
+     Internal restored state dictionary
+    */
+    public let restoredStateData: [String:AnyObject]
```
Cierpliwy: Internal with public visibility?

Source/BluetoothManager.swift
```diff
@@ -394,4 +398,15 @@ public class BluetoothManager {
 				return Observable.just(peripheral)
 		}
 	}
+
+    /**
+     Emits `RestoredState` instance, when state of `BluetoothManager` has been restored
+     - Returns: Observable which emits next events state has been restored
+     */
+    func listenOnRestoredState() -> Observable<RestoredState> {
```
Cierpliwy: There are problems with indentation in this file.

Source/RestoredState.swift
Kacper20: What do you mean by that?

Source/BluetoothManager.swift
Kacper20: How should it look?

Source/RestoredState.swift
literator: In comment you have "internal" while keyword is public.

Source/RestoredState.swift
```diff
@@ -0,0 +1,62 @@
...
+
+    let bluetoothManager: BluetoothManager
```
literator: 1. Consider `unowned` for bluetooth manger
2. Make it public? Maybe user would like to check it out. (Notice that user has access to `restoredStateData` which you pass in `init`)

Kacper20:
Done.

PR #30, opened 2016-06-14
[WIP] CBPeripheralManager support

AndrewSB:


AndrewSB:
Currently the Peripheral object acts as a CBPeripheralDelegate. This causes some issues for the implementation of the CBPeripheralManager, since the peripheral manager needs to be able to pass around Peripheral instances.

Should Peripheral be a generic protocol that can be specialized into something like OtherDevicePeripheral (which would contain the CBPeripheralDelegate implementation) and ThisDevicePeripheral, which would not?

Would love your input before I make this change @Kacper20

Kacper20:
Hi, will check this out by tomorrow. Sorry for the delay :-)

AndrewSB:
Moved to #35

PR #37, opened 2016-06-26
Fix stuck service discovery when looking for different services

xtazz:
Subsequent calls to discoverServices method in Peripheral with different services UUIDs causing it to return an empty sequence instead of success. This happens because after the first call the peripheral already has some discovered services and in the next call the comparison between sets of services identifiers fails if you ask to discover another service

### DIFF
--- a/Source/Peripheral.swift
+++ b/Source/Peripheral.swift
@@ -134,7 +134,7 @@ public class Peripheral {
 					let mapped = discoveredServices.map { Service(peripheral: self, service: $0) }
 					guard let identifiers = serviceUUIDs else { return Observable.just(mapped) }
 					let uuids = discoveredServices.map { $0.uuid }
-					if Set(uuids) == Set(identifiers) {
+					if Set(identifiers).isSubsetOf(Set(uuids)) {
 						return Observable.just(mapped)
 					}
 					return Observable.empty()


Kacper20:
Hi, 
Great catch. Will look at it tomorrow, and probably release it with other bug fixes.

Thanks! :-)

Kacper20:
Hello,
I looked at it with a bit of delay.
I think that your intention was rather to check it on the other way, so instead of: ` Set(identifiers).isSubsetOf(Set(uuids))` it would look like that:
`Set(uuids).isSubsetOf(Set(identifiers))`. So it'll work in a way that if you requested services A,B,C and A was already discovered, so method returns only B,C. 

Thank you very much for discovering this - it's an issue and it'll be resolved today.

Have a great day :-)

xtazz:
Hi! 'identifiers' is a list of identifiers which I pass to the discoverServices, this Set can contain less identifiers than already discovered, so it's a subset of already discovered services Set. But it also should work in other way - if you request more services than you have now, you should only request missing services, so probably my fix should be extended for this case.

Kacper20:
Hello,

We've been thinking about it and i think that we've found optimal solution.
Basically - now we're not taking into consideration services that are given by the delegate method. 
So basically: When you have already discovered service A, and asked about services A, B, C - behavior is that you'll receive all of them, when B, C will be discovered.
Just to be sure - this is behavior you're expecting?

Thank you for feedback :-)

Kacper20:
This will be fixed by #40

xtazz:
You are describing one case, it also should be handled. But in my case I have A,B,C discovered and I request only A in the next call.

Kacper20:
So why do you want to call `discoverServices` while every one of them is already discovered? :-)

xtazz:
Because peripheral contains this information about these services in cache, so this call should succeed in my case

Kacper20:
Ok, that's right - but you could also check it's `services` property and simply read it. Why not? 
But i think that we could implement pre checking - so if every one of requested services is already in cache it'll be wrapped in observable and returned without calling any discover.

xtazz:
You are right, I can do that. But this call shouldn't simply stuck if someone will not do this check.

Kacper20:
I agree with you that this should work - so i'll implement the following behavior: if someone specifies concrete array of identifiers, these will be returned if already in cache.

Thank you for feedback, it's really valuable :)

PR #65, opened 2016-10-05
add write to property functionality to example project

mkulikov:


Example/Example/CharacteristicsController.swift
```diff
@@ -49,16 +51,28 @@ class CharacteristicsController: UIViewController {
             }).addDisposableTo(disposeBag)
     }
 
-    private func showWriteFieldForCharacteristic(characteristic: Characteristic) {
+    fileprivate func showWriteFieldForCharacteristic(characteristic: Characteristic) {
         let valueWriteController = UIAlertController(title: "Write value", message: "Specify value in HEX to write ",
                                                      preferredStyle: .alert)
         valueWriteController.addTextField { textField in
-
+            textField.delegate = self
```
Kacper20: We're setting the delegate right here. But `UIAlertController` has property `textFields` which can give us the necessary data without introducing `Data` property in class.

Example/Example/CharacteristicsController.swift
```diff
@@ -49,16 +51,28 @@ class CharacteristicsController: UIViewController {
...
         }
         valueWriteController.addAction(UIAlertAction(title: "Cancel", style: .cancel, handler: nil))
         valueWriteController.addAction(UIAlertAction(title: "Write", style: .default) { _ in
-            print("")
+            
+            self.writeValueForCharacteristic(data: self.Data!, characteristic: characteristic)
```
Kacper20: I believe here we could read the hex string: `valueWritrController.textFields.first?.text`. Remember to avoid force-unwrapping of optionals - let's stick to the safer versions - `if let` or `guard let`.

Example/Example/CharacteristicsController.swift
```diff
@@ -49,16 +51,28 @@ class CharacteristicsController: UIViewController {
...
+            
             })
+        self.present(valueWriteController, animated: true, completion: nil)
+    }
+    
+    fileprivate func writeValueForCharacteristic(data: String,characteristic: Characteristic) {
+        let _data: NSData = data.dataFromHexadecimalString()!
```
Kacper20: I think that we shouldn't call it `_data`. As far as I know it's not common in Swift. 
Maybe let's do a little bit of a rename: `data` -> `hexadecimalString`. Also, extension on string with this name is not clearest options - i would stick maybe to the static function on `Data` like:
```
extension Data {
static func fromHexString(string: String) -> Data {impl...}
}
```
What do you think? Also - let's avoid unwrapping and printing to the console

Example/Example/CharacteristicsController.swift
```diff
@@ -104,6 +118,14 @@ extension CharacteristicsController: UITableViewDataSource, UITableViewDelegate
             }
             actionSheet.addAction(readValueNotificationAction)
         }
+        
+        if characteristic.properties.contains(.write) {
+            let writeValueNotificationAction = UIAlertAction(title: "Trigger value write", style: .default) { _ in
+                self.showWriteFieldForCharacteristic(characteristic: characteristic)
```
Kacper20: Let's have simple and concise like: `"Write"`

Example/Example/CharacteristicsController.swift
```diff
@@ -124,6 +146,15 @@ extension CharacteristicTableViewCell {
     }
 }
 
+extension CharacteristicsController: UITextFieldDelegate {
+    
+    func textField(_ textField: UITextField, shouldChangeCharactersIn range: NSRange, replacementString string: String) -> Bool {
```
Kacper20: When we'll use `textFields` property like I said above it's not necessary.

Example/Example/StringUtilities.swift
```diff
@@ -0,0 +1,30 @@
...
+extension String {
+    
+    /// Create `NSData` from hexadecimal string representation
+    ///
+    /// This takes a hexadecimal representation and creates a `NSData` object. Note, if the string has any spaces or non-hex characters (e.g. starts with '<' and with a '>'), those are ignored and only hex characters are processed.
+    ///
+    /// - returns: Data represented by this hexadecimal string.
+    
+    func dataFromHexadecimalString() -> NSData? {
+        let data = NSMutableData(capacity: characters.count / 2)
```
Kacper20: I think that since we have `Data` we shouldn't really use `NSMutableData`.

Example/Example/StringUtilities.swift
```diff
@@ -0,0 +1,30 @@
...
+import Foundation
+extension String {
+    
+    /// Create `NSData` from hexadecimal string representation
+    ///
+    /// This takes a hexadecimal representation and creates a `NSData` object. Note, if the string has any spaces or non-hex characters (e.g. starts with '<' and with a '>'), those are ignored and only hex characters are processed.
+    ///
+    /// - returns: Data represented by this hexadecimal string.
+    
+    func dataFromHexadecimalString() -> NSData? {
```
Kacper20: Return `Data`?

Example/Example/StringUtilities.swift
```diff
@@ -0,0 +1,30 @@
...
+        let data = NSMutableData(capacity: characters.count / 2)
+        
+        let regex = try! NSRegularExpression(pattern: "[0-9a-f]{1,2}", options: .caseInsensitive)
```
Kacper20: Let's avoid unwrapping - wrap it in the `if let's` or `guard let's`  and when condition is not meet - return nil. This we'll be a lot safer.

Kacper20:
@mkulikov Any progress on this? :-)

mkulikov:
@Kacper20, yes :) Thank you!

Example/Example/CharacteristicsController.swift
mkulikov: ok)

Example/Example/CharacteristicsController.swift
mkulikov: ok

Example/Example/StringUtilities.swift
mkulikov: ok

Example/Example/StringUtilities.swift
mkulikov: ok

Example/Example/CharacteristicsController.swift
mkulikov: ok

Example/Example/CharacteristicsController.swift
mkulikov: ok

Example/Example/CharacteristicsController.swift
mkulikov: ok

Example/Example/StringUtilities.swift
mkulikov: i use do catch

Example/Example/NSDataUtilities.swift
```diff
@@ -21,4 +21,22 @@ extension Data {
         }
         return NSString(string: hexString) as String
     }
+    
+    // Return Data represented by this hexadecimal string
+    static func fromHexString(string: String) -> Data {
+        var data = Data(capacity: string.characters.count / 2)
+        
+        do {
+            let regex = try! NSRegularExpression(pattern: "[0-9a-f]{1,2}", options: .caseInsensitive)
```
Kacper20: If we're doing `do [...] { catch }` you should just use `try` instead of `try`.

Example/Example/NSDataUtilities.swift
```diff
@@ -21,4 +21,22 @@ extension Data {
...
+            regex.enumerateMatches(in: string, options: [], range: NSMakeRange(0, string.characters.count)) { match, flags, stop in
+                let byteString = (string as NSString).substring(with: match!.range)
+                var num = UInt8(byteString, radix: 16)
+                data.append(&num!, count: 1)
```
Kacper20: Is this `!` necessary?

Example/Example/NSDataUtilities.swift
mkulikov: ok

Example/Example/NSDataUtilities.swift
mkulikov: ok

Kacper20:
Hi there,
This looks good 👍 

Merging it to the master, thanks!

PR #67, opened 2016-10-17
Fixed RxTest.framework reference and minor changes

moogle19:
Fixed the RxTest.framework reference (which was still RxTests and it would not build).

Also fixed some swiftlint warnings (mostly whitespace and line_length) and deactivated missing_docs because it is broken in XCode 8 (see https://github.com/realm/SwiftLint/issues/811).

I also changed the array comparison in RxCharacteristicType and RxServiceType to use starts(with:by:) because when the arrays have the same length and one starts with the other they should be equal.

### DIFF
--- a/.swiftlint.yml
+++ b/.swiftlint.yml
@@ -4,13 +4,13 @@ disabled_rules: # rule identifiers to exclude from running
   - todo #delete it later
 opt_in_rules: # some rules are only opt-in
   - empty_count
-  - missing_docs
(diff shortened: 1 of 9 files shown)


moogle19:
Not sure why the Travis CI build stalled for this PR, but it worked for my branch (https://travis-ci.org/moogle19/RxBluetoothKit/jobs/168215442)

Kacper20:
Merged, thank you very much as always! 👍

PR #68, opened 2016-10-24
Update RxSwift and Nimble

moogle19:
- RxSwift now on 3.0 (finally 🎉 )
- Nimble now on 5.1

I also changed the Cartfile/Podspec version specification to the 'compatible with' operator.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,3 +12,5 @@ before_script:
 script:
         - set -o pipefail && xcodebuild test -verbose -scheme "RxBluetoothKit iOS" -destination "platform=iOS Simulator,OS=10.0,name=iPhone 7 Plus" ONLY_ACTIVE_ARCH=NO | xcpretty
 
+after_success:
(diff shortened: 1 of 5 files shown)


Kacper20:
Hi,
Thanks as always! :-)
Do tou have any idea why Travis check has failed? Looks like Travis problem I suppose, not something with code itself.

moogle19:
Looks like it is a travis bug: https://github.com/travis-ci/travis-ci/issues/4725

I just updated the `.travis.yml` with a `sleep 5` after success, which should fix the error.

I also updated some other stuff in the `.travis.yml` like using the current `Carthage` because the carthage/RxSwift problems described in the previous `.travis.yml` seem resolved.

I also added the `-verbose` flag to xcodebuild, so if the problem (or others) appears again, we may get a better problem description.

Kacper20:
Now it looks all fine.
Merging it to master :-) 

Good work Kevin 👍

PR #69, opened 2016-10-26
Fix BluetoothManager options not being passed to RxCBCentralManager

Orgmir:
I've noticed this warning when using the library:

`[CoreBluetooth] API MISUSE: <private> has no restore identifier but the delegate implements the centralManager:willRestoreState: method. Restoring will not be supported`

Digging a bit, this happens when the `CBCentralManagerOptionRestoreIdentifierKey` is not passed has an option. When actually trying to pass an options dictionary to `BluetoothManager` I noticed this bug.


Kacper20:
You're totally right. Merging it to master right away :) 

Thank you very much for noticing this!

PR #80, opened 2016-11-25
fix bug with type of Characteristic writeValue

mkulikov:


Kacper20:
Thanks for that!

Cheers,

Issue #82, opened 2016-11-29
Support for discovering multiple services & characteristics with the same UUIDs


Kacper20:
This is thing that we're not currently supporting, and it's major issue for now. It would require changing API, so it will land in 4.0 release.
Design of this implementation is currently in progress.

brycejacobs:
What's the holdup of PR #89 ? It doesn't break the surface API from what I can see.

abeintopalo:
I guess this can be closed since PR #109 was merged.

Issue #81, opened 2016-11-28
Support for multiple service instances with same UUID


abeintopalo:
Hello,

First of all I'd like to thank you for your great work to make this awesome library.
I really like the API and it is a pleasure to work with. :)

I have a use case when the peripheral has multiple instances of a BLE service (i.e. multiple service instances with same UUID).
Each service instance have the same set of characteristics.

I have tried the following code to test if RxBluetoothKit can handle this scenario.
The code below supposed to read the value of the same characteristic for each service instances:

```
peripheral.connect()
    .flatMap { $0.discoverServices([serviceId]) }
    .flatMap { Observable.from($0) }
    .flatMap { $0.discoverCharacteristics([characteristicId])}
    .flatMap { Observable.from($0) }
    .flatMap { $0.readValue() }
    .subscribe(onNext: { characteristic in
        guard let data = characteristic.value else { return }
        let dataString = NSString(data: data, encoding: String.Encoding.utf8.rawValue)!
        print("Value: \(dataString)")
    })
```

Unfortunately, the code above prints the characteristic value only for the first service instance.
The value printing code is invoked as many times as many instances of the same service has been discovered but
it seems that the characteristic passed to "subscriber(onNext:)" will be connected only to the first discovered service instance.

I would like to ask if you know any way to get this working when there are multiple service instances with same UUID?
Any help is appreciated.

Kacper20:
Hi there,

Sorry for the late response - we were thinking about it today and unfortunately there's no easy answer to this question. Currently that behavior is not supported by the lib. What we're doing internally in library is manually closing the stream after receiving first service/characteristic that fits given UUID.

We think that it's major issue now, and we're focusing on fixing this thing. We thought about the specific design and unfortunately we're unable to support this in 3.x version - it would break compatibility for other people that are using the library. Also - to make it work right and independent of CoreBluetooth callback implementations we would need to cache `CBService` objects internally(because as we believe in this case it's the only way to distinguish between service instances). Also - we need tochange the Rx stream nature in case of services discover(`discoverService` should never complete I suppose).

I've created an issue for this: #82, and we're working to support this in newest, 4.0 version.

If you need this feature right away we could help you develop some kind of workaround until the 4.0 version will be released.
Thank you very much for feedback,

Kacper

abeintopalo:
Hi Kacper,

Thank you for response. I really appreciate that you are planning to fix this in the next release.
[...]

PR #89, opened 2017-01-13
Fix: reading values when multiple services/characteristics used with …

abeintopalo:
…same UUID (without breaking existing API)

PR is updated!

(See issues: https://github.com/Polidea/RxBluetoothKit/issues/81 and https://github.com/Polidea/RxBluetoothKit/issues/82)

Earlier when discovering services/characteristics UUID comparison was used to
determine for a given service/characteristics if CBPeripheralDelegate callback
was related to a given service/characteristic instance.
It worked only when each service/characteristic had unique UUID.
This approach failed to work when multiple service/characteristic instances used same UUID.
The other issue was that service/characteristics discovery assumed that the result list count
will be the same as the request UUID list count. In practice the result list can be bigger than
the request UUID list because of the possibility of redundant UUIDs.

The commit fixes both issues above:

- Refactored RxServiceType, RxCharacteristicType, RxDescriptorType equality.
Service/Characteristic/Descriptor instances will be considered equal if the the wrapped
CoreBluetooth objects references are equal. This guarantees proper bindings with
CBPeripheralDelegate callbacks with corresponding CoreBluetooth objects.

- Result list count of service/characteristics discoveries filtered based on the original request UUID list

- Some code in unit tests is updated to make sure unit tests are passing

-------

I've created a filter logic that is used to check cached items for discoverServices/discoverIncludedServices/discoverCharacteristics calls. All the unit tests are passing. The template function is filterCachedItems<T>. It would be nice to have additional unit tests for filterCachedItems<T> method but I'm not familiar with QuickSpec way of unit testing. 

I've forced pushed my changes so that you can see easily the diff to the original code.

### DIFF
--- a/Source/Peripheral.swift
+++ b/Source/Peripheral.swift
@@ -376,11 +376,16 @@ public class Peripheral {
     // MARK: Descriptors
     /**
      Function that triggers descriptors discovery for characteristic
+     If all of the descriptors are already discovered - these are returned without doing any underlying Bluetooth operations.
      - Parameter characteristic: `Characteristic` instance for which descriptors should be discovered.
      - Returns: Observable that emits `Next` with array of `Descriptor` instances, once they're discovered.
      Immediately after that `.Complete` is emitted.
      */
     public func discoverDescriptors(for characteristic: Characteristic) -> Observable<[Descriptor]> {
+        if let descriptors = characteristic.descriptors {
+            let resultDescriptors = descriptors.map { Descriptor(descriptor: $0.descriptor, characteristic: characteristic) }
+            return ensureValidPeripheralState(for: .just(resultDescriptors))
+        }
         let observable = peripheral
             .rx_didDiscoverDescriptorsForCharacteristic
             .filter { $0.0 == characteristic.characteristic }
--- a/Source/Peripheral.swift
+++ b/Source/Peripheral.swift
@@ -123,22 +123,20 @@ public class Peripheral {
      Immediately after that `.Complete` is emitted.
      */
     public func discoverServices(_ serviceUUIDs: [CBUUID]?) -> Observable<[Service]> {
-        if let identifiers = serviceUUIDs, let services = self.services?.filter({ identifiers.contains($0.uuid) }), identifiers.count == services.count {
-            return ensureValidPeripheralState(for: .just(services))
+        if let identifiers = serviceUUIDs, !identifiers.isEmpty,
+           let cachedServices = self.services,
+           let filteredServices = Self.filterCachedItems(uuids: serviceUUIDs, items: cachedServices, itemUUID: { $0.uuid }) {
+           return ensureValidPeripheralState(for: .just(filteredServices))
         }
         let observable = peripheral.rx_didDiscoverServices
         .flatMap { (_, error) -> Observable<[Service]> in
-            if let cachedServices = self.services, error == nil {
-                guard let identifiers = serviceUUIDs else { return .just(cachedServices) }
-                let uuids = cachedServices.map { $0.service.uuid }
-                if Set(identifiers).isSubset(of: Set(uuids)) {
-                    let filteredServices = cachedServices
-                        .filter { identifiers.contains($0.uuid)}
-                    return .just(filteredServices)
-                }
-                return .empty()
+            guard let cachedServices = self.services, error == nil else {
+                throw BluetoothError.servicesDiscoveryFailed(self, error)
             }
-            throw BluetoothError.servicesDiscoveryFailed(self, error)
+            if let filteredServices = Self.filterCachedItems(uuids: serviceUUIDs, items: cachedServices, itemUUID: { $0.uuid }) {
+                return .just(filteredServices)
+            }
+            return .empty()
         }
         .take(1)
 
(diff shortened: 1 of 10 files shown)


Kacper20:
Hi,

Thanks for that.
We need to make some evaluation of the code on the real device(it could take a few days because we need to make some mock that has this characteristics). 
We've delayed this change, because of `Swift 3` that doens't allow us to make as pure architecture as we wanted(it misses some major generic features). But after thinking a bit we definitely could support this feature(at least in temporary way) without this changes.

Will dig in the code later today, and prepare some thoughts about it.

Thank you very much once again :) 

Kacper

abeintopalo:
When testing I found this app useful: https://itunes.apple.com/us/app/lightblue-explorer-bluetooth/id557428110?mt=8

It enables creating virtual peripherals on the fly (though with some limitations).

Anyway, when testing the following use cases are important:

- multiple services with same UUID and (having same characteristics with same UUIDs)
- single service having multiple characteristics with same UUID

The lightblue explorer app has some limitations concerning descriptor customizations so from that aspect own test app is a better solution.

Source/Peripheral.swift
```diff
@@ -123,7 +123,7 @@ public class Peripheral {
      Immediately after that `.Complete` is emitted.
      */
     public func discoverServices(_ serviceUUIDs: [CBUUID]?) -> Observable<[Service]> {
-        if let identifiers = serviceUUIDs, let services = self.services?.filter({ identifiers.contains($0.uuid) }), identifiers.count == services.count {
+        if let identifiers = serviceUUIDs, let services = self.services?.filter({ identifiers.contains($0.uuid) }), identifiers.count <= services.count {
             return ensureValidPeripheralState(for: .just(services))
         }
         let observable = peripheral.rx_didDiscoverServices
```
Cierpliwy: From what I can see `self.peripheral.discoverServices(serviceUUIDs)` is always called, even when cache is sufficient for our needs.

Cierpliwy:
```
let uuids = cachedServices.map { $0.service.uuid }
if Set(identifiers).isSubset(of: Set(uuids)) {
```
This code should be reused at the top somehow?

Source/Peripheral.swift
```diff
@@ -165,7 +165,7 @@ public class Peripheral {
     public func discoverIncludedServices(_ includedServiceUUIDs: [CBUUID]?, for service: Service) -> Observable<[Service]> {
```
Cierpliwy: Same things here... Maybe we can reuse above code in better way. Not just copy paste and change few things around.

Source/Peripheral.swift
```diff
@@ -204,7 +204,7 @@ public class Peripheral {
      */
     public func discoverCharacteristics(_ characteristicUUIDs: [CBUUID]?, for service: Service) -> Observable<[Characteristic]> {
         if let identifiers = characteristicUUIDs, let characteristics = service.characteristics?.filter({ identifiers.contains($0.uuid) }),
-            identifiers.count == characteristics.count {
+            identifiers.count <= characteristics.count {
```
Cierpliwy: This check is not correct as stated above.

Source/Peripheral.swift
```diff
@@ -216,8 +216,9 @@ public class Peripheral {
                 }
                 let characteristics = rxCharacteristics.map { Characteristic(characteristic: $0, service: service) }
                 guard let characteristicIdentifiers = characteristicUUIDs else { return .just(characteristics) }
+
                 let filteredCharacteristics = characteristics.filter { characteristicIdentifiers.contains($0.uuid) }
-                if filteredCharacteristics.count == characteristicIdentifiers.count { return .just(filteredCharacteristics) }
+                if characteristicIdentifiers.count <= filteredCharacteristics.count { return .just(filteredCharacteristics) }
```
Cierpliwy: Same here

Source/Peripheral.swift
```diff
@@ -123,7 +123,7 @@ public class Peripheral {
      Immediately after that `.Complete` is emitted.
      */
     public func discoverServices(_ serviceUUIDs: [CBUUID]?) -> Observable<[Service]> {
-        if let identifiers = serviceUUIDs, let services = self.services?.filter({ identifiers.contains($0.uuid) }), identifiers.count == services.count {
+        if let identifiers = serviceUUIDs, let services = self.services?.filter({ identifiers.contains($0.uuid) }), identifiers.count <= services.count {
```
abeintopalo: yes, I was only modifying the existing code == for counts. I didn't want to do more complex refactoring. I've tested my use case an it worked for that but you are right we need more generic solution. I think the basic principle could be is that whenever we request a list of [A,B] the check should examine if this list is subset of the cached one. If yes then we can safely return the filtered list

Source/Peripheral.swift
abeintopalo: I'll check if we can make this a generic solution

abeintopalo:
PR has been update based on review comments. Please, check it.

brycejacobs:
What's the status on this? This is a crucial feature.

abeintopalo:
@brycejacobs The PR is still solid. It seems the guys were busy to respond/include the PR. If you want to use it you can fork my develop branch. At least, that's how I use it in my app.

brycejacobs:
@Cierpliwy Any eta on when this can be merged to cut a release?

brycejacobs:
@abeintopalo My grand master plan is to merge this into the Polidea/react-native-ble-plx#61 

Hopefully by helping to cut a release over there.

abeintopalo:
New PR will be created shortly

Kacper20:
Okay, that's great!
Today we'll have a meeting and we're trying to merge it as soon as possible.

Thank you @abeintopalo

PR #90, opened 2017-01-18
Renaming of `BluetoothManager`. First version of code that allows to share implementation between `Central` and `Peripheral` managers

Kacper20:
I've renamed `BluetoothManager` into `CentralManager.
Also, it appears that the only code that `Central` shares with `Peripheral` in `CoreBluetooth` is the state managing code. I've decided to move this logic to the extension of new protocol `BluetoothStateManager` that by design should be adopted by `Central` and `Peripheral`.

Created empty stub of `PeripheralManager`.

### DIFF
--- a/Source/BluetoothManager.swift
+++ b/Source/BluetoothManager.swift
@@ -193,7 +193,7 @@ public class CentralManager: BluetoothStateManager {
                 return self.ensure(.poweredOn, observable: observable)
             }
     }
-    
+
     // MARK: Peripheral's Connection Management
 
     /**
--- a/Source/BluetoothManager.swift
+++ b/Source/BluetoothManager.swift
@@ -51,10 +51,6 @@ public class CentralManager: BluetoothStateManager {
     /// Implementation of Central Manager
     private let centralManager: RxCentralManagerType
 
-    var stateProvider: BluetoothStateProvider {
-        return centralManager
-    }
-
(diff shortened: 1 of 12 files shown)


Source/BluetoothManager.swift
```diff
@@ -106,7 +110,7 @@ public class BluetoothManager {
      Scans by default are infinite streams of `ScannedPeripheral` structures which need to be stopped by the user. For
      example this can be done by limiting scanning to certain number of peripherals or time:
 
-     bluetoothManager.scanForPeripherals(withServices: nil)
+     CentralManager.scanForPeripherals(withServices: nil)
```
Cierpliwy: I guess it should be lower cased.

Source/BluetoothStateManager.swift
```diff
@@ -0,0 +1,42 @@
...
+    /// Observable which emits state changes of central manager after subscriptions
+    var rx_didUpdateState: Observable<BluetoothState> { get }
+
+    /// Current state of Central Manager
+    var state: BluetoothState { get }
+}
+
+protocol BluetoothStateManager {
+    var stateProvider: BluetoothStateProvider { get }
+}
```
Cierpliwy: Why did you choose another level of indirection behind BluetoothStateManager? Is it really necessary?

Kacper20:
@Cierpliwy fixed it.

PR #91, opened 2017-01-19
Initial hierarchy changes for 4.0 version

Kacper20:


Source/4_0/PeripheralType.swift
```diff
@@ -0,0 +1,47 @@
+import Foundation
+import CoreBluetooth
+import RxSwift
+
+protocol PeripheralType {
```
Cierpliwy: We need to think what will be available for PeripheralType, and what will be available for ConnectedPeripheralType

Source/4_0/CentralManager.swift
```diff
@@ -0,0 +1,47 @@
...
+    public let cbCentralManager: CBCentralManager
+
+    init(cbCentralManager: CBCentralManager) {
+        self.cbCentralManager = cbCentralManager
+    }
+
+    /// Observable which emits state changes of central manager after subscriptions
+    var rx_didUpdateState: Observable<BluetoothState> {
+        return .empty()
+    }
```
Kacper20: I think that this is opened question. Added TODO and we should handle this in future and make some kind of guidelines. We have a couple of reactive values exposed to the users, and probably using underscore is not the best idea in Swift 3. Also, the problem is that `Reactive` is part of `RxCocoa`, not `RxSwift`.

Kacper20: Ok, so I think that these should be written like that, or maybe `_rxDidUpdateState`? Also, I have problem when writing extension to reactive: ```extension Reactive where Base: ManagerType {
    public var mm: Observable<BluetoothState> {
        return .deferred {
            return self.base.rx_didUpdateState.startWith(self.base.state)
        }
    }
}```
It doesn't compile, unfortunately.

Kacper20:
I've merged this PR as it was standing just too long here. Updated with couple of changes based on new Swift 4.0. It's crashing in compile time and I'm investigating this right now :)

PR #97, opened 2017-02-27
Update to RxSwift 3.2 and some cleanup

moogle19:
- Updated RxSwift to 3.2 
- Fixed the swiftlint "vertical alignment" warnings (not sure if this is the preferred formatting?)
- Replaced some flatMap's with map where it made sense (mostly thrown errors)

I also looked through the code and changed some formatting like:
- spacing before/after `:` (e.g `[SomeClass: SomeOtherClass]`)
- remove `else` where it is not needed (because the `if`-part returns)
- replace `if let` with `guard` where it made sense
- replace `map{ SomeClass(something: $0) }` with `map(SomeClass.init)`

### DIFF
--- a/Cartfile.resolved
+++ b/Cartfile.resolved
@@ -1,3 +1,3 @@
 github "Quick/Nimble" "v5.1.1"
 github "Quick/Quick" "v0.10.0"
-github "ReactiveX/RxSwift" "3.1.0"
+github "ReactiveX/RxSwift" "3.2.0"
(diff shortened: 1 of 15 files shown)


Kacper20:
Great work, as always :)
Looks good to me, @Cierpliwy could you also check this?

Cierpliwy:
Looks great to me as well. I think, we can merge it.

PR #109, opened 2017-04-26
Fix: reading values when multiple services/characteristics used with same UUID (without breaking existing API)

abeintopalo:
(See issues: #81 and #82)

Earlier when discovering services/characteristics UUID comparison was used to
determine for a given service/characteristics if CBPeripheralDelegate callback
was related to a given service/characteristic instance.
It worked only when each service/characteristic had unique UUID.
This approach failed to work when multiple service/characteristic instances used same UUID.
The other issue was that service/characteristics discovery assumed that the result list count
will be the same as the request UUID list count. In practice the result list can be bigger than
the request UUID list because of the possibility of redundant UUIDs.

The commit fixes both issues above:

- rebased on top of latest master branch

- Refactored RxServiceType, RxCharacteristicType, RxDescriptorType equality.
Service/Characteristic/Descriptor instances will be considered equal if the the wrapped
CoreBluetooth objects references are equal. This guarantees proper bindings with
CBPeripheralDelegate callbacks with corresponding CoreBluetooth objects.

- Result list count of service/characteristics discoveries filtered based on the original request UUID list

- Some code in unit tests is updated to make sure unit tests are passing

- added a filter logic that is used to check cached items for discoverServices/discoverIncludedServices/discoverCharacteristics calls. All the unit tests are passing. The template function is filterCachedItems.

- Added caching for discovered descriptors

- Fixed Xcode 8.3.2 warning(s)

- Additional minor fixes

BTW: we are testing this with real HW and it works.

### DIFF
--- a/Source/Peripheral.swift
+++ b/Source/Peripheral.swift
@@ -375,11 +375,16 @@ public class Peripheral {
     // MARK: Descriptors
     /**
      Function that triggers descriptors discovery for characteristic
+     If all of the descriptors are already discovered - these are returned without doing any underlying Bluetooth operations.
      - Parameter characteristic: `Characteristic` instance for which descriptors should be discovered.
      - Returns: Observable that emits `Next` with array of `Descriptor` instances, once they're discovered.
      Immediately after that `.Complete` is emitted.
      */
     public func discoverDescriptors(for characteristic: Characteristic) -> Observable<[Descriptor]> {
+        if let descriptors = characteristic.descriptors {
+            let resultDescriptors = descriptors.map { Descriptor(descriptor: $0.descriptor, characteristic: characteristic) }
+            return ensureValidPeripheralState(for: .just(resultDescriptors))
+        }
         let observable = peripheral
             .rx_didDiscoverDescriptorsForCharacteristic
             .filter { $0.0 == characteristic.characteristic }
--- a/Source/Peripheral.swift
+++ b/Source/Peripheral.swift
@@ -123,19 +123,17 @@ public class Peripheral {
      Immediately after that `.Complete` is emitted.
      */
     public func discoverServices(_ serviceUUIDs: [CBUUID]?) -> Observable<[Service]> {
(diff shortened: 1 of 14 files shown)


Source/Peripheral.swift
```diff
@@ -543,3 +547,25 @@ extension Peripheral: Equatable { }
...
+    /**
+     Filters an item list based on the provided UUID list. Only items returned whose UUID matches an item in the provided UUID list.
+     Each UUID should have at least one item matching in the items list. Otherwise the result is nil.
+
+     - uuids: a UUID list or nil
+     - items: items to be filtered
+     - itemUUID: closure to extract the UUID from the items
+     - Returns: the filtered item list
+    */
+    static func filterCachedItems<T>(uuids: [CBUUID]?, items: [T], itemUUID: (T) -> CBUUID) -> [T]? {
```
Kacper20: Maybe instead of T which is unconstrained we could constrain in to some protocol(names are not thought through)
```swift
protocol UUIDIdentifiable {
  var uuid: CBUUID { get }
}
extension CBCharacteristic: UUIDIdentifiable { }
extension CBService: UUIDIdentifiable { }
```
What about making it global? It's not associated with the `Peripheral` in any way.
Thanks to this we could have one less parameter in function call and we avoid closures.
Then interface of `filterUUIDItems` method would look like:
```swift
func filterUUIDItems<T: UUIDIdentifiable>(uuids: [CBUUID]?, items: [T]) -> [T]? {
```

abeintopalo: @Kacper20 Thank you for your feedback. I think your suggestion is a pretty good idea. I'm gonna implement the changes based on your comments and force push the updated PR.

abeintopalo:
@Kacper20 updated the PR based on your comments. I kind of liked the name `UUIDIdentifiable`. 
:)

abeintopalo:
I've tested unit tests and they are passing. Could you restart Travis build? For some reason it failed.

Kacper20:
Looks great. Merging it to master. Hopefully we'll be able to release a new version by Friday.
Thank you very much!

Best regards,
Kacper

abeintopalo:
Glad I could help. It is an awesome library. :)

PR #115, opened 2017-05-23
Add option to dynamically enable logging

Cierpliwy:
- Currently only CoreBluetooth delegates are logged as most basic entry points for BLE debugging.
- Log levels where adjusted to RxAndroidBle library for better compatibility.
- `print()` function was used for faster logging.
- `@autoclosure` was used to execute message only when `currentLogLevel >= messageLogLevel`.
- Added missing `objectId` for Central Manager to be able to distinguish logs from multiple managers.

### DIFF
--- a/Source/Logging.swift
+++ b/Source/Logging.swift
@@ -52,7 +52,7 @@ public class RxBluetoothKitLog {
     }
 
     fileprivate static func log(with logLevel: LogLevel, message: @autoclosure () -> String) {
-        if currentLogLevel.rawValue >= logLevel.rawValue {
+        if currentLogLevel >= logLevel {
             print(tag(with: logLevel), message())
         }
     }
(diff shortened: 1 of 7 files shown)


Source/Logging.swift
```diff
@@ -0,0 +1,125 @@
...
+        let time = Date().timeIntervalSinceReferenceDate
+        return prefix + String(format: "%02.0f:%02.0f:%02.0f.%03.f]:",
+                               floor(time / 3600.0).truncatingRemainder(dividingBy: 24),
+                               floor(time / 60.0).truncatingRemainder(dividingBy: 60),
+                               floor(time).truncatingRemainder(dividingBy: 60),
+                               floor(time * 1000).truncatingRemainder(dividingBy: 1000))
+    }
+
+    fileprivate static func log(with logLevel: LogLevel, message: @autoclosure () -> String) {
+        if currentLogLevel.rawValue >= logLevel.rawValue {
```
Kacper20: We could write equatable for `LogLevel`

PR #126, opened 2017-07-30
Restricted peripheral state to connected when read is called

Kacper20:
Read operation now has to happen during connected state. In other case error is thrown to the stream.
It's implemented as in other places in `Peripheral` object.

R: @Cierpliwy

### DIFF
--- a/Source/Peripheral.swift
+++ b/Source/Peripheral.swift
@@ -339,8 +339,8 @@ public class Peripheral {
      - returns: Observable which emits `Next` with Characteristic that state was changed. Immediately after `.Complete`
      is emitted
      */
-    public func setNotifyValue(_ enabled: Bool, for characteristic: Characteristic)
-        -> Observable<Characteristic> {
+    public func setNotifyValue(_ enabled: Bool,
+                               for characteristic: Characteristic) -> Observable<Characteristic> {
         let observable = peripheral
             .rx_didUpdateNotificationStateForCharacteristic
             .filter { $0.0 == characteristic.characteristic }
@@ -474,10 +474,9 @@ public class Peripheral {
         )
     }
 
-    func ensureValidPeripheralStateAndCallIfSucceeded<T>(
-        for observable: Observable<T>,
-        postSubscriptionCall: @escaping () -> Void
-        ) -> Observable<T> {
+    func ensureValidPeripheralStateAndCallIfSucceeded<T>(for observable: Observable<T>,
+                                                         postSubscriptionCall call: @escaping () -> Void
+                                                        ) -> Observable<T> {
         let operation = Observable<T>.deferred {
             call()
             return .empty()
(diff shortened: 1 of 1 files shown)


Source/Peripheral.swift
```diff
@@ -326,7 +326,8 @@ public class Peripheral {
      */
     public func readValue(for characteristic: Characteristic) -> Observable<Characteristic> {
         return .create { observer in
-            let disposable = self.monitorValueUpdate(for: characteristic).take(1).subscribe(observer)
+            let disposable = self.ensureValidPeripheralState(for: self.monitorValueUpdate(for: characteristic).take(1))
+                .subscribe(observer)
             self.peripheral.readValue(for: characteristic.characteristic)
```
Cierpliwy: I think we should make sure that readValue will not be called as a side effect in situation where subscribe call above detects any invalid state which prohibits BLE operations. `readValue` function should be called after `monitorValueUpdate` subscription.

Kacper20: Yeah, you're right. We definitely should change the way it is handled in this & other places as it's only partial solution. I will think about a way and update PR accordingly

Kacper20:
@Cierpliwy Thanks for the review!
I've updated the PR according to what we've talked about.

Source/Peripheral.swift
```diff
@@ -148,11 +148,10 @@ public class Peripheral {
...
         .take(1)
 
-        return .create { observer in
-            let disposable = self.ensureValidPeripheralState(for: observable).subscribe(observer)
-            self.peripheral.discoverServices(serviceUUIDs)
-            return disposable
-        }
+        return ensureValidStateAndCallIfNecessary(
+            for: observable,
+            call: { self.peripheral.discoverServices(serviceUUIDs) }
```
Cierpliwy: I'm not sure how to point out that this call is always called _after_ observable subscription.

Source/Peripheral.swift
```diff
@@ -342,23 +339,23 @@ public class Peripheral {
...
-                .rx_didUpdateNotificationStateForCharacteristic
-                .filter { $0.0 == characteristic.characteristic }
-                .take(1)
-                .map { (_, error) -> Characteristic in
-                    if let error = error {
-                        throw BluetoothError.characteristicNotifyChangeFailed(characteristic, error)
-                    }
-                    return characteristic
+    public func setNotifyValue(
+        _ enabled: Bool,
```
Cierpliwy: Indentation is strange here.

Source/Peripheral.swift
```diff
@@ -471,12 +468,22 @@ public class Peripheral {
...
-            self.peripheral.writeValue(data, for: descriptor.descriptor)
-            return disposable
+        let monitorWrite = self.monitorWrite(for: descriptor).take(1)
+        return ensureValidStateAndCallIfNecessary(
+            for: monitorWrite,
+            call: { self.peripheral.writeValue(data, for: descriptor.descriptor) }
+        )
+    }
+
+    func ensureValidStateAndCallIfNecessary<T>(
```
Cierpliwy: I would like to point out that this is special case for `ensureValidPeripheralState`
and also that later call is executed only when subscription of observable is successful.

`ensureValidPeripheralState(for: callPostSubscription: )`

Issue #42, opened 2016-08-08
Access to underlying CBPeripheral


mobiletoly:
Let's say I already have an instance of RxBlueoothKit::Peripheral class available. How to get an access to underlying CBPeripheral instance? I have third-party library that performs Device Firmware Update (DFU) :
https://github.com/NordicSemiconductor/IOS-DFU-Library/tree/master/documentation

and it asks for reference to CBCentralManager and CBPeripheral, so would be great if I could just get it from already existing instance of RxBlueoothKit::Peripheral.

Thank you, your library is outstanding.

Kacper20:
Hi there,
Basically, you have reference to `RxBluetoothKit::Peripheral`. This instance has access to the `peripheral` property which is `RxPeripheralType` instance. This is a protocol, which allows us to test our implementation. We've built on it Fake objects, aswell as real `RxCBPeripheral` object. This object has property `let peripheral: CBPeripheral`. Unfortunately, because thanks to this `RxPeripheralType` protocol, usage assumes that you shouldn't access `RxCBPeripheral` because it's implementation detail. It's class with internal access so it's not really possible as of now.

One solution that comes to my mind quickly is to fork library and add `public` scope to two places:
```
public class RxCBPeripheral: RxPeripheralType {

	public let peripheral: CBPeripheral
	[...]
}`
Then, assuming that your `Peripheral` instance is named `peripheral` you'd write it like:
`let cbPeripheral = (peripheral.peripheral as! RxCBPeripheral).peripheral`

Which looks ugly, but should work properly.
We'll talk about your problem internally, because it's interesting case and we should take such into consideration. 
We'll let you know soon about our thoughts. 

Thanks for help in improving library!

Kacper20:
Hi,

Have you managed to deal with this problem somehow? 

Thanks!

jwhitley:
Hi all, I work with @mobiletoly now, and have gone through a few iterations of adding accessors for `CBCentralManager` and `CBPeripheral` to support our integration with Nordic's firmware update library.  Please see [this branch in our fork of RxBluetoothKit](https://github.com/LoopDevices/RxBluetoothKit/tree/jwhitley/cb-access) for the current work, and [its diff vs. master](https://github.com/Polidea/RxBluetoothKit/compare/master...LoopDevices:jwhitley/cb-access).

[...]

PR #129, opened 2017-08-04
Public access to `CBPeripheral` instance given

Kacper20:
It's the change that we've talked about @Cierpliwy.
I wonder about one more thing - should we give user access to `Peripheral` constructor?
```
    init(peripheral: CBPeripheral) {
        self.peripheral = peripheral
        self.internalDelegate = RxCBPeripheral.getInternalPeripheralDelegateRef(cbPeripheral: peripheral)
    }
```
So that user, after using `CBPeripheral` needed by another lib is able to reclaim control over peripheral to our library?
They easily could do it by calling `peripheralsWithIdentifiers` on `BluetoothManager` so I'm not sure whether it's needed

### DIFF
--- a/Source/BluetoothManager.swift
+++ b/Source/BluetoothManager.swift
@@ -66,6 +66,10 @@ public class BluetoothManager {
         return centralManager.objectId
     }
 
+    public var manager: CBCentralManager {
+        return centralManager.centralManager
+    }
+
     // MARK: Initialization
 
     /**
--- a/Source/Peripheral.swift
+++ b/Source/Peripheral.swift
@@ -31,7 +31,7 @@ import CoreBluetooth
  allowing to talk to peripheral like discovering characteristics, services and all of the read/write calls.
  */
 public class Peripheral {
-    let manager: BluetoothManager
+    public let manager: BluetoothManager
 
     init(manager: BluetoothManager, peripheral: RxPeripheralType) {
         self.manager = manager
--- a/Source/Peripheral.swift
+++ b/Source/Peripheral.swift
@@ -60,6 +60,13 @@ public class Peripheral {
     }
 
     /**
+     Underlying `CBPeripheral` instance
+     */
+    public var cbPeripheral: CBPeripheral {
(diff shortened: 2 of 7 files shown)


jwhitley:
This looks good, but I'll request one further addition along the same lines: an accessor for the `CBCentralManager`. See [my comment on #42](https://github.com/Polidea/RxBluetoothKit/issues/42#issuecomment-262621755) for more details, as well as the CBCentralManager changes in the [diff of my fork](https://github.com/Polidea/RxBluetoothKit/compare/master...LoopDevices:jwhitley/cb-access).

Kacper20:
@jwhitley Thanks for pointing this out! That's reasonable and I'll expose this, too.

Kacper

Kacper20:
@Cierpliwy @jwhitley I've updated the PR. Could you take a look?

jwhitley:
Reviewing my branch's diffs, I think we also need to be able to retrieve the CBCentralManager associated with a Peripheral.  @Kacper20 I'm going to build our app against your branch and report back ASAP to check that everything's covered.

jwhitley:
@Kacper20 Confirmed.  Nordic's library requires both the `CBPeripheral` and the `CBCentralManager` to initiate DFU.  These are pulled from a `Peripheral`, since that's what we've got at the point where DFU starts.  I've created and annotated a gist that illustrates how we use  RxBluetoothKit with Nordic's DFU library.  Note the comments on lines 48-50:

https://gist.github.com/5925f89ceba2520a6b322b3dbdeaf732

Kacper20:
Thanks @jwhitley,
I'm lookin' into it.

Kacper20:
@jwhitley I've updated the PR. Now you could take `BluetoothManager` from `Peripheral` and do everything is required by Nordic.

I'm merging it now, thanks for feedback once again!

jwhitley:
@Kacper20 Just confirmed the 3.1 release as working for us.  Great work, thanks!

Kacper20:
I'm glad it works! Thanks!

PR #146, opened 2017-10-02
Fix Travis for Swift 4

moogle19:
Changed the Xcode version from 8.3 to 9 and simulator from 10 to 11 for Swift 4 support.


Kacper20:
Great, thanks!

Issue #120, opened 2017-06-19
[5.0] Use `Single` RxSwift unit where possible


moogle19:
Since [RxSwift 3.3](https://github.com/ReactiveX/RxSwift/releases/tag/3.3.0) there is the `Single` type, which emit only one Element or an error. 
I think this should be use for everything that just emits one value, like `readValue` and `writeValue`.
Because it is a breaking change, it should probably be added in 4.0

Kacper20:
That's a good idea.
We're in progress with 4.0 right now and we'll definitely use it.

Thanks!

Kacper20:
Implemented as a part of #147 on `5.0-dev` branch

PR #147, opened 2017-10-02
Switch from `Observable` to  `Single` where applicable

moogle19:
Changed `Observable` to `Single` where only one element is emitted. (see #120 )

Currently it is just done by using the `asSingle()` method of the observables.
The tests then use the `asObservable()` method, to make them still work.
In the future this should be done more elegantly.

I am not sure if the `swift4` branch is the right target, but the `4.0-dev` wasn't update for quite a long time.

### DIFF
--- a/Source/Peripheral+Convenience.swift
+++ b/Source/Peripheral+Convenience.swift
@@ -42,6 +42,7 @@ extension Peripheral {
                 return .just(service)
             } else {
                 return self.discoverServices([identifier.uuid])
+                    .asObservable()
                     .flatMap { Observable.from($0) }
             }
         }
@@ -65,6 +66,7 @@ extension Peripheral {
                         return .just(characteristic)
                     }
                     return service.discoverCharacteristics([identifier.uuid])
+                        .asObservable()
                         .flatMap { Observable.from($0) }
                 }
         }
@@ -87,6 +89,7 @@ extension Peripheral {
                         return .just(descriptor)
                     }
                     return characteristic.discoverDescriptors()
+                        .asObservable()
                         .flatMap { Observable.from($0) }
                         .filter { $0.uuid == identifier.uuid }
                         .take(1)
--- a/Source/Peripheral+Convenience.swift
+++ b/Source/Peripheral+Convenience.swift
@@ -35,7 +35,7 @@ extension Peripheral {
      - Returns: Observation which emits `Next` event, when specified service has been found.
      Immediately after that `.Complete` is emitted.
      */
-    public func service(with identifier: ServiceIdentifier) -> Observable<Service> {
+    public func service(with identifier: ServiceIdentifier) -> Single<Service> {
         return .deferred {
             if let services = self.services,
                 let service = services.first(where: { $0.uuid == identifier.uuid  }) {
@@ -44,6 +44,7 @@ extension Peripheral {
                 return self.discoverServices([identifier.uuid])
                     .asObservable()
                     .flatMap { Observable.from($0) }
+                    .asSingle()
             }
         }
     }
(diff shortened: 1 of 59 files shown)


Source/Peripheral+Convenience.swift
```diff
@@ -136,6 +143,7 @@ extension Peripheral {
      */
     public func monitorValueUpdate(for identifier: CharacteristicIdentifier) -> Observable<Characteristic> {
         return characteristic(with: identifier)
+            .asObservable()
```
Kacper20: Why do we need `asObservable()` here?

Source/Peripheral+Convenience.swift
```diff
@@ -103,6 +109,7 @@ extension Peripheral {
     public func monitorWrite(for identifier: CharacteristicIdentifier)
         -> Observable<Characteristic> {
         return characteristic(with: identifier)
+            .asObservable()
```
Kacper20: Why do we need `asObservable()` here?

Source/Peripheral+Convenience.swift
```diff
@@ -78,18 +82,20 @@ extension Peripheral {
...
         return .deferred {
             return self.characteristic(with: identifier.characteristic)
-                .flatMap { characteristic -> Observable<Descriptor> in
+                .flatMap { characteristic -> Single<Descriptor> in
                     if let descriptors = characteristic.descriptors,
                         let descriptor = descriptors.first(where: { $0.uuid == identifier.uuid }) {
                         return .just(descriptor)
                     }
                     return characteristic.discoverDescriptors()
+                        .asObservable()
```
Kacper20: Why do we need `asObservable()` here?

Source/Peripheral+Convenience.swift
moogle19: This one is necessary because `monitorValueUpdate` is an infinite stream and therefore must be an `Observable`

Kacper20: Right, makes sense. Thanks!

Source/Peripheral+Convenience.swift
moogle19: This can be replace with something like this:
```
return self.discoverServices([identifier.uuid])
                    .map {
                        if let service = $0.first {
                            return service
                        }
                        throw RxError.noElements
                    }
```

or changed to a `Maybe`.
I am not sure what the expected behaviour is, when the Service UUID is not present in the Peripheral. Should it throw an error (`Single`) or just complete (`Maybe`)?

Kacper20:
@moogle19 I have one more requests — could you update documentation on these functions accordingly?

Source/Peripheral+Convenience.swift
moogle19: @Kacper20 @Cierpliwy Any suggestions regarding this?

moogle19:
@Kacper20 Yes, I'm on it.

Source/Peripheral+Convenience.swift
Kacper20: We're gonna discuss it today as it touches the expected behavior and we don't want to break logic for clients of the library. Thanks for the patience! :)

Kacper20:
Thanks! Any comments @Cierpliwy? After thinking we would like to release it under `5.0` version which will be breaking one. We want `4.0` to be a safe transition from iOS 10 to iOS 11 ( and macOS accordingly) SDK. This will make it easier for our users to continue use library without much effort.

Kacper20:
I will create 5.0 branch and merge it tomorrow. Thanks for the patience.

Kacper20:
Thanks for rebasing :) This will make my life soo much easier! 💯

Kacper20:
Thanks! Great work as always! :)

PR #171, opened 2017-10-23
Resolves #162 — Architecture cleanup and unit tests removal

### LINKED ISSUE #162: [5.0] Cleanup of library from the old architecture
As part of 5.0 release we want to simplify the library by changing it’s architecture. This will allow us to remove any code duplication that is part of the current release of the library
# Change design
Current hierarchy for every `CoreBluetooth` object looks as follows:
- Main object that user of the library is interacting with ie. `Peripheral`
- Protocol that exposes delegate method callbacks & methods that perform Bluetooth related operations (scan, connection) ie. `RxPeripheralType`.
- Concrete implementation of the protocol that is using `CoreBluetooth` objects: `RxCBPeripheral`. These are injected to the main objects and allow for the normal use of the library
- Fakes that are implementing the protocols ie. `FakePeripheral`. We’re injecting them into main objects for the purpose of unit tests.

We would like to get rid of the last 3 layers when implementing the cleanup.  After the currently described issue will be implemented tests will not be a part of the library and will come back a bit later. We believe that the new approach to testing the library will be much easier and allow to make changes and API additions faster. It should also make tests way more concise than today.
As a result of this task:
- All protocols, concrete implementations and fakes should be removed
- Tests related to the `Bluetooth` stack hierarchy should be removed

Then, all main, user facing objects will directly use it’s `CB` counterparts and call the methods on them. To receive delegate callbacks, `BluetoothManager` and `Peripheral` will contain delegate wrappers object. First one will conform to the `CBCentralManagerDelegate` protocol and the second one to the `CBPeripheralDelegate` protocol. Both will expose delegate callbacks via observables to the main objects.

# How it affects the current users
External API of the `BluetoothManager`, `Peripheral`, `Service`, `Characteristic`, `Descriptor` and any of the dependent objects will not be changed. Therefore it is not considered a breaking change.

### LINKED ISSUE #163: [5.0] Prevention from stealing delegates between Peripheral instances, way to (re-)attach delegate.
# Stealing delegates inside of our library
Let’s imagine the stream of `Peripheral` instances which looks as follows in the code: `Observable<Peripheral`. The source of values is Bluetooth scan. Underneath whenever we’re receiving new `CBPeripheral` on delegate callback there is new `Peripheral` instance created. Inside creation of `Peripheral` we’re creating new delegate proxy object. Let’s make an assumption tthat stream emits duplicate peripherals (like it currently when [this particular](https://developer.apple.com/documentation/corebluetooth/cbcentralmanagerscanoptionallowduplicateskey?language=swift) option is set to `true`).
Then, every time that duplicate is emitted it has different memory address and all of the predecessors are unusable right now (because the delegate on this particular `CBPeripheral` that is bound to our objects is overwritten by the newest instance). This behavior may be confusing to the users of the library. To mitigate this behavior we propose to introduce caching layer inside the `BluetoothManager`. If `Peripheral` is currently in cache we would like to get it from there and not create another instance. It’ll have proper delegates set. There are many possible implementations — it’s up to the contributor taking this issue to propose something more detailed. One edge-case to have in mind is to always check whether `CBPeripheral` discovered has our own proxy object set. If not we should probably reattach delegate to it (as it would be done when creating `Peripheral` object). Since now it’ll be cached we should be making this additional check.

# Stealing delegates outside of our library
User of our library has access to the underlying `CBPeripheral` and `CBCentralManager` objects. This objects could be used to access them in a non-reactive way or pass to the external library like [Nordic-DFU](https://github.com/NordicSemiconductor/IOS-DFU-Library). While doing this the certain thing is that delegate will be overwritten. Currently our library doesn’t give any way to reattach delegate to our objects like `BluetoothManager` and `Peripheral` which are managing `CBCentralManager` and `CBPeripheral` objects.

This proposal suggests adding function `attach()` to these two classes. Implementation of the function would reassign the delegate of managed objects to our reactive delegate proxies.

# How it affects current users
It’ll make using our library along with other ones (like Nordic firmware update) more pleasant. API changes are clearly additive.

# Tasks
- [x] Stealing delegates inside our library
- [ ] Stealing delegate outside of our library

Kacper20:
According to the design of #162 I've done work on removing old architecture and simplifying the code. 
Resolves #162 
# Breaking changes
I've decided that we could change names under which `CBPeripheral` and `CBCentralManager` are exposed. They will probably be removed in another PR so it shouldn't be a big concern.


Source/BluetoothManager.swift
```diff
@@ -58,23 +59,19 @@ public class BluetoothManager {
     /// Unique identifier of an object. Should be removed in 4.0
     @available(*, deprecated)
     public var objectId: UInt {
-        return centralManager.objectId
-    }
-
-    public var manager: CBCentralManager {
-        return centralManager.centralManager
+        return UInt(bitPattern: ObjectIdentifier(centralManager))
```
Cierpliwy: If centralManager is public I guess we don't need this function anymore.

Source/BluetoothManager.swift
```diff
@@ -191,14 +188,14 @@ public class BluetoothManager {
...
-            return self.centralManager.rx_didUpdateState.startWith(self.centralManager.state)
+            return self.delegateWrapper.rx_didUpdateState.startWith(self.state)
         }
     }
 
     /// Current state of `BluetoothManager` instance described by `BluetoothState` which is equivalent to [CBManagerState](https://developer.apple.com/reference/corebluetooth/cbmanager/1648600-state).
     /// - returns: Current state of `BluetoothManager` as `BluetoothState`.
     public var state: BluetoothState {
-        return centralManager.state
+        return BluetoothState(rawValue: centralManager.state.rawValue) ?? .unsupported
```
Cierpliwy: Can't we return centralManager.state directly?

Source/CBCentralManagerDelegateWrapper.swift
```diff
@@ -0,0 +1,100 @@
+// The MIT License (MIT)
+//
+// Copyright (c) 2016 Polidea
```
Cierpliwy: 2017

Source/CBCentralManagerDelegateWrapper.swift
```diff
@@ -0,0 +1,100 @@
...
+
+    private let didUpdateStateSubject = PublishSubject<BluetoothState>()
+    private let willRestoreStateSubject = ReplaySubject<[String: Any]>.create(bufferSize: 1)
+    private let didDiscoverPeripheralSubject = PublishSubject<(CBPeripheral, [String: Any], NSNumber)>()
+    private let didConnectPerihperalSubject = PublishSubject<CBPeripheral>()
+    private let didFailToConnectPeripheralSubject = PublishSubject<(CBPeripheral, Error?)>()
+    private let didDisconnectPeripheral = PublishSubject<(CBPeripheral, Error?)>()
+
+    @objc func centralManagerDidUpdateState(_ central: CBCentralManager) {
+        guard let bleState = BluetoothState(rawValue: central.state.rawValue) else { return }
```
Cierpliwy: Again can we return state directly?

Source/CBCentralManagerDelegateWrapper.swift
```diff
@@ -0,0 +1,100 @@
...
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+// SOFTWARE.
+
+import Foundation
+import CoreBluetooth
+import RxSwift
+
+@objc class CBCentralManagerDelegateWrapper: NSObject, CBCentralManagerDelegate {
```
Cierpliwy: I just wonder if this class could be autogenerated. We could use part of generation code to create FakePeripherals in the same manner.

Source/Characteristic.swift
```diff
@@ -40,7 +40,7 @@ public class Characteristic {
     /// Unique identifier of an object. Should be removed in 4.0
     @available(*, deprecated)
     public var objectId: UInt {
-        return characteristic.objectId
+        return UInt(bitPattern: ObjectIdentifier(characteristic))
```
Cierpliwy: Maybe we should allow to to get CBCharacteristic from Characteristic? If we will be sticking to objectId API, maybe it's better to just return ObjectIdentifier(...) as it is Equatable and Hashable.

Source/Descriptor.swift
```diff
@@ -29,15 +29,15 @@ import RxSwift
...
+    let descriptor: CBDescriptor
 
     /// Characteristic to which this descriptor belongs.
     public let characteristic: Characteristic
 
     /// Unique identifier of an object. Should be removed in 4.0
     @available(*, deprecated)
     public var objectId: UInt {
-        return descriptor.objectId
+        return UInt(bitPattern: ObjectIdentifier(descriptor))
```
Cierpliwy: Same as above.

Source/Peripheral.swift
```diff
@@ -72,13 +69,13 @@ public class Peripheral {
...
     public var identifier: UUID {
-        return peripheral.identifier
+        return peripheral.value(forKey: "identifier") as! NSUUID as UUID
     }
 
     /// Unique identifier of `Peripheral` object instance. Should be removed in 4.0
     @available(*, deprecated)
     public var objectId: UInt {
-        return peripheral.objectId
+        return UInt(bitPattern: ObjectIdentifier(peripheral))
```
Cierpliwy: Check comments above.

Source/Service.swift
```diff
@@ -40,7 +40,7 @@ public class Service {
     /// Unique identifier of an object. Should be removed in 4.0
     @available(*, deprecated)
     public var objectId: UInt {
-        return service.objectId
+        return UInt(bitPattern: ObjectIdentifier(service))
```
Cierpliwy: Same here.

Source/CBCentralManagerDelegateWrapper.swift
Kacper20: Yep, surely we will try do to this as a part of another task

Source/CBCentralManagerDelegateWrapper.swift
Kacper20: It's still necessary.

Issue #169, opened 2017-10-23
[5.0] L2CAPChannels and writeWithoutResponse enhancements in iOS 11 & macOS 13


Kacper20:
In [What’s New in CoreBluetooth, 712, WWDC 2017](https://developer.apple.com/videos/play/wwdc2017/712/) session Apple has announced couple of new APIs as a part of iOS 11 & macOS 13. We would like to support it in our reactive wrappers.

# Design
There were 2 main API changes introduced: L2CAPChannels capabilities in `Central` modes of the CoreBluetooth and enhancements in write without response. L2CAPChannels change should be also a part of #166 task.
## L2CAPChannels
There are two stages of this task: Central and Peripheral implementation. 
### Central Mode
In central mode (currently supported by our library) `CBPeripheral` exposes `openL2CAPChannel` method. When L2CAPChannel is opened method `peripheral(CBPeripheral, didOpen: CBL2CAPChannel?, error:Error?)` is called on `CBPeripheralDelegate`. 
What we should do it to expose `openL2CAPChannel` method on `Peripheral` that would return `Observable<L2CAPChannel>` (with name subject to change). It would do standard things, visibel in other similar methods like `connect`: observing delegate callbacks and emit Error / Next accordingly.
Proper errors should be introduced as a part of `BluetoothError` structure

We should also allow user to monitor openings of `L2CAPChannel`. To do this we should expose function like `func monitorL2CAPChannelOpenings -> Observable<CBL2CAPChannel`. It would connect to [this](https://developer.apple.com/documentation/corebluetooth/cbperipheralmanagerdelegate/2880154-peripheralmanager) delegate method. We should probably provide function like `publishAndMonitorOpeningsOfL2CAPChannel`. Basically the design should be analogous to the setting notification as enabled and monitoring notifications on `CBCharacteristic`.

### L2CAPChannel
It’s a wrapper around `CBL2CAPChannel` CoreBluetooth object. It should only be created if we would decide to wrap `InputStream` and `OutputStream` properties of `CBL2CAPChannel` into reactive wrappers. If it’ll not be the case (and it *probably* shouldn’t be) — all of the interface methods should simply return directly `CBL2CAPChannel` objects.

## Write Without Response Enhancements
To mark this task as a complete there should be:
- `var canSendWriteWithoutResponse: Bool` exposed as a read-only variable on `Peripheral` instances. It should just read the value from the underlying `CBPeripheral` instance and return it to the user.
[...]

PR #174, opened 2017-10-25
Write without response advancements 

Kacper20:
I've exposed the API call & delegate method for checking readiness of peripheral to write without response. It's integrated into our current write method, too.
Partially addresses #169

### DIFF
--- a/Source/CBPeripheralDelegateWrapper.swift
+++ b/Source/CBPeripheralDelegateWrapper.swift
@@ -54,6 +54,9 @@ import RxSwift
     var rx_didWriteValueForDescriptor: Observable<(CBDescriptor, Error?)> {
         return peripheralDidWriteValueForDescriptorSubject
     }
+    var rx_peripheralReadyToSendWriteWithoutResponse: Observable<Void> {
+        return peripheralIsReadyToSendWriteWithoutResponseSubject
+    }
 
     private let peripheralDidUpdateNameSubject = PublishSubject<String?>()
     private let peripheralDidModifyServicesSubject = PublishSubject<([CBService])>()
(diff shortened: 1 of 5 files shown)


Source/CBPeripheralDelegateWrapper.swift
```diff
@@ -194,4 +198,11 @@ import RxSwift
             """)
         peripheralDidWriteValueForDescriptorSubject.onNext((descriptor, error))
     }
+
+    func peripheralIsReady(toSendWriteWithoutResponse peripheral: CBPeripheral) {
+        peripheralIsReadyToSendWriteWithoutResponseSubject.onNext(())
+        RxBluetoothKitLog.d("""
```
Cierpliwy: I would prefer log to be called before emitting next event.

Source/Peripheral.swift
```diff
@@ -245,23 +252,38 @@ public class Peripheral {
     /// - `withResponse` -  Observable emits `Next` with `Characteristic` instance write was confirmed without any errors.
     /// Immediately after that `Complete` is called. If any problem has happened, errors are emitted.
     /// - `withoutResponse` - Observable emits `Next` with `Characteristic` instance once write was called.
-    /// Immediately after that `.Complete` is called. Result of this call is not checked, so as a user you are not sure if everything completed successfully. Errors are not emitted
+    /// Immediately after that `.Complete` is called. Result of this call is not checked, so as a user you are not sure if everything completed successfully. Errors are not emitted. It ensures that peripheral is ready to write without
```
Cierpliwy: Quite long comment. Maybe we could split it earlier?

Source/Peripheral.swift
```diff
@@ -245,23 +252,38 @@ public class Peripheral {
...
+            return Observable<Characteristic>.deferred { [weak self] in
+                guard let strongSelf = self else { throw BluetoothError.destroyed }
+                let canWrite = strongSelf.monitorWriteWithoutResponseReadiness()
+                    .map { _ in true }
+                    .startWith(strongSelf.canSendWriteWithoutResponse)
+                    .filter { $0 }
+                    .take(1)
+                let observable = writeOperationPerformingAndListeningObservable(Observable.just(characteristic))
+                return canWrite
+                    .flatMap { _ in observable }
```
Cierpliwy: I would write above without introducing tmp variables like: canWrite and observable.

Kacper20: True, I was fighting the type system. Will change it.

PR #175, opened 2017-10-26
Resolves #170 - Return error on monitor disconnection

### LINKED ISSUE #170: [5.0] `monitorDisconnection` on `Peripheral` instance should return the error to the user
Currently method signature for `monitorDisconnection` looks as follows:
`public func monitorDisconnection(for peripheral: Peripheral) -> Observable<Peripheral>`
Delegate callback that is used to implement this contains additional information regarding the error which is currently ommited by our library. 
# Design
Ideally we would change the interface to the following one:
```
typealias DisconnectionReason = Error?
public func monitorDisconnection(for peripheral: Peripheral) -> Observable<(Peripheral, DisconnectionReason)>
```
# How it affects current users
It’s a source breaking change and should be a part of 5.0 release.

bartoszstelmaszuk:
Resolves #170 so the library returns error on monitorDisconnection to the user

### DIFF
--- a/Source/BluetoothError.swift
+++ b/Source/BluetoothError.swift
@@ -38,6 +38,7 @@ public enum BluetoothError: Error {
     // Peripheral
     case peripheralConnectionFailed(Peripheral, Error?)
     case peripheralDisconnected(Peripheral, Error?)
+    case peripheralDisconnectionFailed(Peripheral, Error?)
     case peripheralRSSIReadFailed(Peripheral, Error?)
     // Services
     case servicesDiscoveryFailed(Peripheral, Error?)
@@ -77,6 +78,8 @@ extension BluetoothError: CustomStringConvertible {
         case let .peripheralConnectionFailed(_, err):
             return "Connection error has occured: \(err?.localizedDescription ?? "-")"
         case let .peripheralDisconnected(_, err):
+            return "Connection error has occured: \(err?.localizedDescription ?? "-")"
+        case let .peripheralDisconnectionFailed(_, err):
             return "Connection error has occured: \(err?.localizedDescription ?? "-")"
         case let .peripheralRSSIReadFailed(_, err):
             return "RSSI read failed : \(err?.localizedDescription ?? "-")"
--- a/Source/BluetoothError.swift
+++ b/Source/BluetoothError.swift
@@ -38,7 +38,6 @@ public enum BluetoothError: Error {
     // Peripheral
     case peripheralConnectionFailed(Peripheral, Error?)
     case peripheralDisconnected(Peripheral, Error?)
-    case peripheralDisconnectionFailed(Peripheral, Error?)
     case peripheralRSSIReadFailed(Peripheral, Error?)
     // Services
     case servicesDiscoveryFailed(Peripheral, Error?)
@@ -78,8 +77,6 @@ extension BluetoothError: CustomStringConvertible {
         case let .peripheralConnectionFailed(_, err):
             return "Connection error has occured: \(err?.localizedDescription ?? "-")"
         case let .peripheralDisconnected(_, err):
-            return "Connection error has occured: \(err?.localizedDescription ?? "-")"
-        case let .peripheralDisconnectionFailed(_, err):
             return "Connection error has occured: \(err?.localizedDescription ?? "-")"
         case let .peripheralRSSIReadFailed(_, err):
             return "RSSI read failed : \(err?.localizedDescription ?? "-")"
--- a/Source/BluetoothManager.swift
+++ b/Source/BluetoothManager.swift
@@ -255,7 +255,7 @@ public class BluetoothManager {
     /// - parameter peripheral: The `Peripheral` to which the `BluetoothManager` is either trying to
     /// connect or has already connected.
     /// - returns: `Single` which emits next event when peripheral successfully cancelled connection.
-    public func cancelPeripheralConnection(_ peripheral: Peripheral) -> Observable<(Peripheral, BluetoothError?)> {
+    public func cancelPeripheralConnection(_ peripheral: Peripheral) -> Single<(Peripheral, BluetoothError?)> {
         let observable = Observable<(Peripheral, BluetoothError?)>.create { [weak self] observer in
             guard let strongSelf = self else {
                 observer.onError(BluetoothError.destroyed)
@@ -265,7 +265,7 @@ public class BluetoothManager {
             strongSelf.centralManager.cancelPeripheralConnection(peripheral.peripheral)
             return disposable
         }
-        return ensure(.poweredOn, observable: observable)
+        return ensure(.poweredOn, observable: observable).asSingle()
     }
 
     // MARK: Retrieving Lists of Peripherals
--- a/Source/BluetoothManager.swift
+++ b/Source/BluetoothManager.swift
@@ -255,8 +255,8 @@ public class BluetoothManager {
     /// - parameter peripheral: The `Peripheral` to which the `BluetoothManager` is either trying to
     /// connect or has already connected.
     /// - returns: `Single` which emits next event when peripheral successfully cancelled connection.
-    public func cancelPeripheralConnection(_ peripheral: Peripheral) -> Single<(Peripheral, BluetoothError?)> {
-        let observable = Observable<(Peripheral, BluetoothError?)>.create { [weak self] observer in
+    public func cancelPeripheralConnection(_ peripheral: Peripheral) -> Single<(Peripheral)> {
+        let observable = Observable<(Peripheral, DisconnectionReason)>.create { [weak self] observer in
             guard let strongSelf = self else {
                 observer.onError(BluetoothError.destroyed)
                 return Disposables.create()
@@ -265,7 +265,7 @@ public class BluetoothManager {
             strongSelf.centralManager.cancelPeripheralConnection(peripheral.peripheral)
             return disposable
         }
-        return ensure(.poweredOn, observable: observable).asSingle()
+      return ensure(.poweredOn, observable: observable).asSingle().map { $0.0 }
     }
 
     // MARK: Retrieving Lists of Peripherals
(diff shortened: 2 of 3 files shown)


Source/BluetoothManager.swift
```diff
@@ -255,8 +255,8 @@ public class BluetoothManager {
     /// - parameter peripheral: The `Peripheral` to which the `BluetoothManager` is either trying to
     /// connect or has already connected.
     /// - returns: `Single` which emits next event when peripheral successfully cancelled connection.
-    public func cancelPeripheralConnection(_ peripheral: Peripheral) -> Single<Peripheral> {
-        let observable = Observable<Peripheral>.create { [weak self] observer in
+    public func cancelPeripheralConnection(_ peripheral: Peripheral) -> Observable<(Peripheral, BluetoothError?)> {
```
siejkowski: Maybe we could introduce a wrapper type to indicate that the peripheral might be returned without error?

Source/BluetoothManager.swift
```diff
@@ -265,7 +265,7 @@ public class BluetoothManager {
             strongSelf.centralManager.cancelPeripheralConnection(peripheral.peripheral)
             return disposable
         }
-        return ensure(.poweredOn, observable: observable).asSingle()
```
siejkowski: Is there a reason for dropping the `Single` here? Can we emit multiple events now?

Source/BluetoothManager.swift
```diff
@@ -342,22 +342,29 @@ public class BluetoothManager {
...
     /// Emits `Peripheral` instance when it's disconnected.
     /// - Parameter peripheral: `Peripheral` which is monitored for disconnection.
-    /// - Returns: Observable which emits next events when `peripheral` was disconnected.
-    public func monitorDisconnection(for peripheral: Peripheral) -> Observable<Peripheral> {
-        return monitorPeripheral(on: delegateWrapper.rx_didDisconnectPeripheral.map { $0.0 }, peripheral: peripheral)
+    /// - Returns: Observable which emits next events when `peripheral` was disconnected or error event when occured.
+    public func monitorDisconnection(for peripheral: Peripheral) -> Observable<(Peripheral, BluetoothError?)> {
+        return monitorPeripheral(
+            on: delegateWrapper.rx_didDisconnectPeripheral.map { (_, error) in
+                throw BluetoothError.peripheralDisconnectionFailed(peripheral, error)
```
siejkowski: I might not read it correctly, but is there a reason for why you're ignoring the peripheral returned from `peripheralAction` observable here? It seems to me that the logic deciding which instance should be returned in the error should be kept in one place and `monitorPeripheral` is better place in my opinion.

Source/BluetoothManager.swift
Cierpliwy: What do you mean by that? Optional type indicates that there may or may not be an error related to disconnection. How would you name your wrapper type? Will it be an enum?

Source/BluetoothManager.swift
```diff
@@ -342,22 +342,29 @@ public class BluetoothManager {
     /// - Parameter peripheral: `Peripheral` which is monitored for connection.
     /// - Returns: Observable which emits next events when `peripheral` was connected.
     public func monitorConnection(for peripheral: Peripheral) -> Observable<Peripheral> {
-        return monitorPeripheral(on: delegateWrapper.rx_didConnectPeripheral, peripheral: peripheral)
+        return monitorPeripheral(
+            on: delegateWrapper.rx_didConnectPeripheral.map { ($0, nil) },
+            peripheral: peripheral
+        ).map { $0.0 }
```
Cierpliwy: Are we creating tuple to just drop it at the end?

Source/BluetoothError.swift
```diff
@@ -78,6 +79,8 @@ extension BluetoothError: CustomStringConvertible {
             return "Connection error has occured: \(err?.localizedDescription ?? "-")"
         case let .peripheralDisconnected(_, err):
             return "Connection error has occured: \(err?.localizedDescription ?? "-")"
+        case let .peripheralDisconnectionFailed(_, err):
+            return "Connection error has occured: \(err?.localizedDescription ?? "-")"
```
Cierpliwy: According to documentation disconnection cannot fail. So there shouldn't be any error related to that.

Source/BluetoothManager.swift
Cierpliwy: This code is not behaving correctly. We shouldn't throw error there.

Source/BluetoothManager.swift
```diff
@@ -342,22 +342,29 @@ public class BluetoothManager {
...
+        return monitorPeripheral(
+            on: delegateWrapper.rx_didDisconnectPeripheral.map { (_, error) in
+                throw BluetoothError.peripheralDisconnectionFailed(peripheral, error)
+            }, peripheral: peripheral
+        )
     }
 
-    func monitorPeripheral(on peripheralAction: Observable<CBPeripheral>, peripheral: Peripheral)
-        -> Observable<Peripheral> {
+    func monitorPeripheral(on peripheralAction: Observable<(CBPeripheral, BluetoothError?)>, peripheral: Peripheral)
```
Cierpliwy: I'm not sure if this function is worth to share it between connected/disconnected events.

Source/Peripheral.swift
```diff
@@ -92,7 +92,7 @@ public class Peripheral {
     /// that physical connection will be closed immediately as well and all pending commands will not be executed.
     ///
     /// - returns: `Single` which emits next event when peripheral successfully cancelled connection.
-    public func cancelConnection() -> Single<Peripheral> {
+    public func cancelConnection() -> Observable<(Peripheral, BluetoothError?)> {
```
Cierpliwy: Why observable?

Source/BluetoothManager.swift
bartoszstelmaszuk: Looking through library i found that in most cases a tuple is used instead of wrapper type. I decided to follow the current style. Don't you think that optionality on `BluetoothError` indicates enough that the observable can return without error?

Source/BluetoothManager.swift
bartoszstelmaszuk: Good catch! Already repairing it.

Source/BluetoothManager.swift
Cierpliwy: in Observable<T>, T is successful object, so for me using tuple there is OK.

Source/BluetoothManager.swift
Kacper20: +1. In the issue I said that error should be emitted in the resulting observable next events, not as an error on stream.

Source/BluetoothError.swift
Kacper20: +1

Source/BluetoothManager.swift
bartoszstelmaszuk: corrected throwing. It should be mapped as you mentioned.

Source/BluetoothManager.swift
bartoszstelmaszuk: Yes, it is to use `monitorPeripheral` function in both connection and disconnection functions.

Source/BluetoothManager.swift
bartoszstelmaszuk: in my opinion it is. If we would like to separate it instead of two easy maps we would have to implement logic of filtering and mapping on stream  of `pheripheralAction` with ensuring poweredOn state in both function `monitorConnection` and `monitorDisconnection`

Source/Peripheral.swift
```diff
@@ -92,7 +92,7 @@ public class Peripheral {
     /// that physical connection will be closed immediately as well and all pending commands will not be executed.
     ///
     /// - returns: `Single` which emits next event when peripheral successfully cancelled connection.
-    public func cancelConnection() -> Single<Peripheral> {
+    public func cancelConnection() -> Single<(Peripheral)> {
```
Kacper20: We don't need parentheses here.

Source/BluetoothManager.swift
```diff
@@ -265,7 +265,7 @@ public class BluetoothManager {
             strongSelf.centralManager.cancelPeripheralConnection(peripheral.peripheral)
             return disposable
         }
-        return ensure(.poweredOn, observable: observable).asSingle()
+      return ensure(.poweredOn, observable: observable).asSingle().map { $0.0 }
```
Kacper20: I would move map to the new line for better readability

Source/BluetoothManager.swift
```diff
@@ -342,22 +342,29 @@ public class BluetoothManager {
...
+            peripheral: peripheral
+        ).map { $0.0 }
     }
 
     /// Emits `Peripheral` instance when it's disconnected.
     /// - Parameter peripheral: `Peripheral` which is monitored for disconnection.
-    /// - Returns: Observable which emits next events when `peripheral` was disconnected.
-    public func monitorDisconnection(for peripheral: Peripheral) -> Observable<Peripheral> {
-        return monitorPeripheral(on: delegateWrapper.rx_didDisconnectPeripheral.map { $0.0 }, peripheral: peripheral)
+    /// - Returns: Observable which emits next events when `peripheral` was disconnected and error which is
```
Kacper20: I would say: ```Returns: Observable which emits next events when `Peripheral` instance was disconnected. It provides optional error which may contain more information about the cause of the disconnection if it wasn't the `cancelConnection` call```

Source/BluetoothManager.swift
```diff
@@ -342,22 +342,29 @@ public class BluetoothManager {
...
+    /// a disconnection reason when it was caused by 'peripheral' otherwise it's nil.
+    public typealias DisconnectionReason = Error?
```
Kacper20: type alias should be above the documentation comments. Right now docs refer to type alias not to a function. Additionaly it should be `DisconnectionReason = Error`. Then in interface user sees immediately that it's optional — without looking at type alias definition.

Source/BluetoothManager.swift
```diff
@@ -342,22 +342,29 @@ public class BluetoothManager {
...
+    public func monitorDisconnection(for peripheral: Peripheral) -> Observable<(Peripheral, DisconnectionReason)> {
+        return monitorPeripheral(
+            on: delegateWrapper.rx_didDisconnectPeripheral, peripheral: peripheral
+        )
     }
 
-    func monitorPeripheral(on peripheralAction: Observable<CBPeripheral>, peripheral: Peripheral)
-        -> Observable<Peripheral> {
+    func monitorPeripheral(on peripheralAction: Observable<(CBPeripheral, DisconnectionReason)>, peripheral: Peripheral)
+            -> Observable<(Peripheral, Error?)> {
```
Kacper20: We could also use DisconnectionReason here instead of an Error

Source/BluetoothManager.swift
Kacper20: I'm not convinced to this — It's right now kind of artificial way to do this generalization. Could we find some other approach to implement this? This tuple creation reason is complicated and deserves comment at least. The best would be to find other way or just drop helper function.

PR #176, opened 2017-10-26
Added L2CAP channel support for central mode

scibas:
Central mode for #169- L2CAPChannels and writeWithoutResponse enhancements in iOS 11 & macOS 13

### DIFF
--- a/Source/CBPeripheralDelegateWrapper.swift
+++ b/Source/CBPeripheralDelegateWrapper.swift
@@ -57,6 +57,12 @@ import RxSwift
     var rx_peripheralReadyToSendWriteWithoutResponse: Observable<Void> {
         return peripheralIsReadyToSendWriteWithoutResponseSubject
     }
+    
+    @available(OSX 10.13, iOS 11, *)
+    var rx_didOpenL2CAPChannel: Observable<(CBL2CAPChannel?, Error?)> {
+        return peripheralDidOpenL2CAPChannelSubject
+            .map {($0.0 as? CBL2CAPChannel, $0.1)}
+    }
 
     private let peripheralDidUpdateNameSubject = PublishSubject<String?>()
     private let peripheralDidModifyServicesSubject = PublishSubject<([CBService])>()
@@ -73,6 +79,7 @@ import RxSwift
     private let peripheralDidUpdateValueForDescriptorSubject = PublishSubject<(CBDescriptor, Error?)>()
     private let peripheralDidWriteValueForDescriptorSubject = PublishSubject<(CBDescriptor, Error?)>()
     private let peripheralIsReadyToSendWriteWithoutResponseSubject = PublishSubject<Void>()
+    private let peripheralDidOpenL2CAPChannelSubject = PublishSubject<(Any?, Error?)>()
 
     @objc func peripheralDidUpdateName(_ peripheral: CBPeripheral) {
         RxBluetoothKitLog.d("""
@@ -204,5 +211,14 @@ import RxSwift
             \(peripheral.logDescription) peripheralIsReady(toSendWriteWithoutResponse)
             """)
         peripheralIsReadyToSendWriteWithoutResponseSubject.onNext(())
+    }
+    
+    @available(OSX 10.13, iOS 11, *)
+    @objc func peripheral(_ peripheral: CBPeripheral, didOpen channel: CBL2CAPChannel?, error: Error?) {
+        RxBluetoothKitLog.d("""
+            \(peripheral.logDescription) didOpenL2CAPChannel(for:\(peripheral.logDescription),
+            error: \(String(describing: error)))
+            """)
+        peripheralDidOpenL2CAPChannelSubject.onNext((channel, error))
     }
 }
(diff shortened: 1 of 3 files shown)


Kacper20:
LGTM! Thanks! @Cierpliwy?

Source/Peripheral.swift
```diff
@@ -533,6 +533,30 @@ public class Peripheral {
     public func monitorWriteWithoutResponseReadiness() -> Observable<Void> {
         return delegateWrapper.rx_peripheralReadyToSendWriteWithoutResponse
     }
+    
+    /// Function that allow to open LL2CAP channel for `Peripheral` instance.
```
bartoszstelmaszuk: should be "L2CAP"

Source/CBPeripheralDelegateWrapper.swift
```diff
@@ -73,6 +79,7 @@ import RxSwift
     private let peripheralDidUpdateValueForDescriptorSubject = PublishSubject<(CBDescriptor, Error?)>()
     private let peripheralDidWriteValueForDescriptorSubject = PublishSubject<(CBDescriptor, Error?)>()
     private let peripheralIsReadyToSendWriteWithoutResponseSubject = PublishSubject<Void>()
+    private let peripheralDidOpenL2CAPChannelSubject = PublishSubject<(Any?, Error?)>()
```
bartoszstelmaszuk: why `Any?` not `CBL2CAPChannel` ?

Source/CBPeripheralDelegateWrapper.swift
```diff
@@ -57,6 +57,12 @@ import RxSwift
     var rx_peripheralReadyToSendWriteWithoutResponse: Observable<Void> {
         return peripheralIsReadyToSendWriteWithoutResponseSubject
     }
+    
+    @available(OSX 10.13, iOS 11, *)
+    var rx_didOpenL2CAPChannel: Observable<(CBL2CAPChannel?, Error?)> {
+        return peripheralDidOpenL2CAPChannelSubject
+            .map {($0.0 as? CBL2CAPChannel, $0.1)}
```
bartoszstelmaszuk: Do we need to map it? I think that we get from `peripheral: didOpen: error:` an CBL2CAPChannel so if `PublishSubject` will receive in `onNext` this type and not `Any?` we could get rid off this map.

Source/Peripheral.swift
scibas: Will fix

moogle19:
Should `CBL2CAPChannel` be exposed directly or wrapped into a custom class (e.g. `L2CAPChannel`) which implements the `InputStream` and `OutputStream` as `Observable`?

Source/Peripheral.swift
```diff
@@ -533,6 +533,32 @@ public class Peripheral {
...
+    /// For more information, please refer to
+    /// [What’s New in CoreBluetooth, 712, WWDC 2017](https://developer.apple.com/videos/play/wwdc2017/712/)
+    ///
+    /// - parameter PSM: `PSM` (Protocol/Service Multiplexer) of the channel
+    /// - returns: `Single` that emits `CBL2CAPChannel` when channel has opened
+    /// - since: iOS 11, tvOS 11, watchOS 4
+    #if os(iOS)
+        @available(iOS 11, tvOS 11, watchOS 4, *)
+        public func openL2CAPChannel(PSM: CBL2CAPPSM) -> Single<CBL2CAPChannel> {
+            let observable = delegateWrapper.rx_didOpenL2CAPChannel
```
Cierpliwy: take(1)?

scibas:
@moogle19 I've spoke about wrapping `CBL2CAPChannel` with @Kacper20 and we decide not to do so. The main reason is that `CBL2CAPChannel` is fully synchronous class and there is a small chance that API with streams as `Observable`s will be useful for anyone.

Source/Peripheral.swift
```diff
@@ -533,6 +533,32 @@ public class Peripheral {
...
+                .flatMap { [weak self] (channel, error) -> Observable<CBL2CAPChannel> in
+                    guard let strongSelf = self else { throw BluetoothError.destroyed }
+                    if let channel = channel, error != nil {
+                        return .just(channel)
+                    } else {
+                        throw BluetoothError.openingL2CAPChannelFailed(strongSelf, error)
+                    }
+            }
+    
+            return ensureValidPeripheralStateAndCallIfSucceeded(for: observable, postSubscriptionCall: { [weak self] in
```
Cierpliwy: I guess there should version of this function which accepts `Single`s. Name is too long I guess...

Source/Peripheral.swift
scibas: Is `take(1)` really needed in this case? `flatmap` in next line emits `.just` or  throws an `error` so the stream will be closed anyway.

moogle19: If `rx_didOpenL2CAPChannel` emits **n** values, then the code in `flatMap` is called **n** times.
So I think `take(1)` is necessary or else `asSingle()` will emit an error.

scibas: Yeah you guys right, will fix.

Issue #166, opened 2017-10-23
Support for CBPeripheralManager


Kacper20:
Our library currently only supports `CoreBluetooth` central mode. 
New object would be named `PeripheralManager` (with consistent naming to `CentralManager`)

# Design
Since `CBCentralManager` and `CBPeripheralManager` both inherit from `CBManager` which consist `state` property it would be appropriate to map the same relationship in our hierarchy in order to avoid code duplication. It could be named `BaseManager` and should take `CBManager` instance in initializer. This class would expose functions that operate on state, that currently belong to the `BluetoothManager`, like continuous access to `BluetoothState`.
Then, both new `CentralManager` and `PeripheralManager` would inherit from `BaseManager`.

Design of `PeripheralManager` should be analogous to the current `CentralManager` (After changes from #162 . One that implements it should perform pairing of methods from `CBPeripheralManager` and the delegates callbacks. Then, every method should be wrapped in an reactive way. Delegate proxy for the `CBPeripheralManagerDelegate` should also be created.

In [What’s New in CoreBluetooth, 712, WWDC 2017](https://developer.apple.com/videos/play/wwdc2017/712/) session Apple has announced couple of new APIs as a part of iOS 11 & macOS 13. We would like to support it in our reactive wrappers.

## L2CAPChannels
There were 2 main API changes introduced: L2CAPChannels capabilities in `Central` modes of the CoreBluetooth and enhancements in write without response. L2CAPChannels change should be also a part of #166 task.

`CBPeripheralManager` instance contains two methods for publishing and unpublishing channels:
- `publishL2CAPChannel(withEncryption: Bool)`
- `unpublishL2CAPChannel(CBL2CAPPSM)`
These two have delegate methods that inform about the success of each operation. Exposed interface should be as follows:
- `func publishL2CAPChannel(withEncryption: Bool) -> Observable<CBL2CAPPSM>`
- `func unpublishL2CAPChannel(CBL2CAPPSM) -> Observable<CBL2CAPPSM>`
Implementation is pretty standard —> connection to the delegate methods and performing function call on `CBPeripheralManager` instance. Proper errors should be introduced as a part of `BluetoothError` structure.

[...]

PR #177, opened 2017-10-26
Peripheral mode — basic setup, few methods implemented

Kacper20:
In this PR:

I've setup a new class for `CBPeripheralManager` reactive wrapper. I've renamed the current `BluetoothManager` into `CentralManager` which probably makes more sense. 
Implemented sharing of basic things like state management and helper function between two managers.
For `PeripheralManager` full wrappers around the delegates are written.
Also implemented initialisers, basic properties and `startAdvertising` method. It still needs some polishing like documentation. I will add this later today. 

Partially addresses #166 
More of the interface for `PeripheralManager` will be added in subsequent PRs — it's done this way to break it down into smaller pieces, more manageable to review them.

Please let me know what you think!

### DIFF
--- a/Source/PeripheralManager.swift
+++ b/Source/PeripheralManager.swift
@@ -1,0 +1,66 @@
+// The MIT License (MIT)
+//
+// Copyright (c) 2016 Polidea
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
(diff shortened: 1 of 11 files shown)


Source/PeripheralManager.swift
```diff
@@ -0,0 +1,119 @@
...
+                    }
+                    return .just(())
+            }.subscribe(observer)
+            let advertisementChangesDisposable = advertisementData
+                .asObservable()
+                .skip(1)
+                .subscribe(onNext: { [weak self] advertisement in
+                    self?.peripheralManager.startAdvertising(advertisement)
+                })
+            strongSelf.peripheralManager.startAdvertising(advertisementData.value)
```
Cierpliwy: Remove this line and skip(1) from above?

Kacper20: Good catch! Thanks!

PR #180, opened 2017-11-15
Can accessible to 'internal' protection level

kingiol:
I use manager.retrievePeripherals, then get Peripheral
I want to modify Peripheral to ScannedPeripheral
so should use the public init on ScannedPeripheral

### DIFF
--- a/Source/ScannedPeripheral.swift
+++ b/Source/ScannedPeripheral.swift
@@ -36,7 +36,7 @@ public class ScannedPeripheral {
     /// Scanned peripheral's RSSI value.
     public let rssi: NSNumber
 
-    init(peripheral: Peripheral, advertisementData: AdvertisementData, rssi: NSNumber) {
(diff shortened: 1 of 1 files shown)


Kacper20:
Hi there,
What’s the reason to convert `Peripheral` to `ScannedPeripheral`? What will you pass as rssi and advertisementData in constructor?

kingiol:
In my app, I use the ScannedPeriperhal.advertisementData.localName to get real BLE name.
App function's parameter type is all ScannedPeriperhal.

Kacper20:
Yeah but when you’re using `retrievePeripheral` call you don’t receive things like advertisementData and RSSI. You’re forced to pass the values yourself. What do you want to pass to this constructor of `ScannedPeripheral` in this case? I’m asking out of curiosity — right now I don’t see it but maybe I’m overlooking something.

Thanks in advance!

paweljaneczek:
Closeing for now. @kingiol If you feel that it is still needed just answer @Kacper20 questions so we will open PR again ;)

PR #186, opened 2017-11-29
Simplify scanning implementation

Cierpliwy:
Scanning observable is currently able to delay scan subscription when other one is in progress and due to different parameters cannot be shared. Previously subscription queue was subscribing and disposing on specified thread. This caused an issue where next scan subscription could happen before disposal of previous one even when from user perspective order was different.

I decided to remove subscribing on different thread as library is not guaranteed to be thread safe at all. This commit fixes case when start -> stop scan cycle performed quickly causes observable to not emit scan results.

### DIFF
--- a/Source/BluetoothManager.swift
+++ b/Source/BluetoothManager.swift
@@ -100,11 +100,10 @@ public class BluetoothManager {
     /// ```
     ///
     /// If different scan is currently in progress and peripherals needed by a user can be discovered by it, new scan is
-    /// shared. Otherwise scan is queued on thread specified in `init(centralManager:queueScheduler:)` and will be executed
-    /// when other scans finished with complete/error event or were unsubscribed.
-    /// As a result you will receive `ScannedPeripheral` which contains `Peripheral` object, `AdvertisementData` and
-    /// peripheral's RSSI registered during discovery. You can then `connectToPeripheral(_:options:)` and do other
-    /// operations.
+    /// shared. Otherwise scan is queued and will be executed when other scans finished with complete/error event or
+    /// were unsubscribed. As a result you will receive `ScannedPeripheral` which contains `Peripheral` object,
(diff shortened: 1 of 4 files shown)


Source/BluetoothManager.swift
```diff
@@ -70,11 +70,8 @@ public class BluetoothManager {
     /// Creates new `BluetoothManager` instance with specified implementation of `RxCentralManagerType` protocol which will be
     /// used by this class. Most of a time `RxCBCentralManager` should be chosen by the user.
     /// - parameter centralManager: Implementation of `RxCentralManagerType` protocol used by this class.
-    /// - parameter queueScheduler: Scheduler on which all serialised operations are executed (such as scans). By default main thread is used.
-    init(centralManager: RxCentralManagerType,
-         queueScheduler: SchedulerType = ConcurrentMainScheduler.instance) {
+    init(centralManager: RxCentralManagerType) {
```
dariuszseweryn: Update comment in line 103 of the current file regarding the init method.

Source/Observable+QueueSubscribeOn.swift
```diff
@@ -25,51 +25,39 @@ import RxSwift
...
         queue.append(observable)
         if execute {
             // Observable is scheduled immidiately
-            queue.first?.delayedSubscribe(on: scheduler)
+            queue.first?.delayedSubscribe()
         }
     }
 
     func unsubscribe(observable: DelayedObservableType) {
-        lock.lock(); defer { lock.unlock() }
```
dariuszseweryn: Are you sure that unsubscribing two scans from different threads will not change the `index` of the `observable`? Unlikely but seems possible

dariuszseweryn: Queueing seems to need synchronization too

Source/BluetoothManager.swift
Cierpliwy: Done :)

Issue #168, opened 2017-10-23
[5.0] New testing infrastructure


Kacper20:
As a part of #162 all of the tests and the current testing infrastructure will be removed. Rationale behind this decision is written as a part of the #162. 
We would like to perform testing based on the code generation. Basically every implementation class we currently have should have code generated that will use stubs instead of concrete `CB` objects. 
This way we will not have code that we want to duplicate and our architecture and code we write will be independent of testing infrastructure.
# Design
We should use something like `Sourcery` to perform codegen of the new classes. All of the tests for the library should be rewritten. 
Earlier we’ve used `Quick` and `Nimble` for testing. It was partially unfortunate choice — we have lot’s of tests that should be generic. For example: We should check bluetooth error propagation for literally every call that involves any Bluetooth operation. `Quick` at the time of writing tests in 2016 didn’t have enough support for generic testing methods. 

As part of this task there should re-evaluation of `Quick` capabilities. If it haven’t changed since 2016 we advise to use standard `XCTests` to avoid duplication of code. 

# Does it affects current users
No.

paweljaneczek:
Done in #200

PR #196, opened 2017-12-20
Added CB types for mocking

paweljaneczek:
This PR is a proposal to how we want handle testing in library #168. For couple days I've been investigating how to make best approach which will allow make as less problems as possible in future development.

From my perspective best approach would be to:
- create new protocol types which are conformed by `CoreBluetooth` types (made in this PR)
- make use of that protocol types in `BluetoothManager`, `Peripheral`, `Service`, `Characteristic` classes (made in this PR)
- manually create mocks of that types in tests
- write tests with injecting those mocks to `BluetoothManager` and `Peripheral` instances

Some details from my research:
- I've been investigation option with not creating those protocols and doing automated generation of new  `BluetoothManager` test class that will have different entry parameter (I think that it was @Kacper20 and @Cierpliwy approach) but doing this with [Sourcery](https://github.com/krzysztofzablocki/Sourcery) library would be very hard, complicated and could lead to many issues related to that code generation. 
- I've been testing automated generation of mocks using [Sourcery](https://github.com/krzysztofzablocki/Sourcery). There is example template on [sourcery github](https://github.com/krzysztofzablocki/Sourcery/blob/master/Templates/Templates/AutoMockable.stencil), but it need a lot of changes to make work with our types (e.g. it doesn't handle methods with same names, which is possible in swift). It is possible to create template that works with our types but it will need to be very complicated and will be hard to maintain it in future. 
- I've also found library [SwiftyMocky](https://github.com/MakeAWishFoundation/SwiftyMocky) which is using Sourcery for creating their own implementation of mock. It looks like it is working. Problem here is that this library is too much for our needs. It has it's own api wrapper for stubbing, veryfing of mock which may works for some simple cases but may also cause problems in more complicated code. I think that depending so much on that library could cause problems in future development
- We have only 2 types (`CBCentralManager` and `CBPeripheral`)  so I think adding here additional libraries for mock generation would be too much for our needs. That's why my proposal is to add those mocks manually.

I know that you had something different approach in mind, so if you have any doubts or question just write it here ;)

BREAKING CHANGE: I've made change to public wrapped CB type properties, e.g. before if user wanted to get `CBService` he called `Service.service`, now he will need change this to `Service.getService()`.

### DIFF
--- a/Source/CBCentralManagerType.swift
+++ b/Source/CBCentralManagerType.swift
@@ -1,0 +1,65 @@
+// The MIT License (MIT)
+//
+// Copyright (c) 2017 Polidea
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in all
+// copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+// SOFTWARE.
+
+import Foundation
+import CoreBluetooth
+
+/// CBCentralManagerType is protocol that has exact same methods and variables as CBCentralManager
+/// and CBCentralManager implements that protocol.
+/// The only difference here is bluetoothState - it is because in iOS 10 there was type change of
+/// state property and we can not create 2 variables with different type and same name.
+/// It is needed for creating mocks of CBCentralManager in tests.
+
+protocol CBCentralManagerType: class {
+    var delegate: CBCentralManagerDelegate? { get set }
+    var bluetoothState: BluetoothState { get }
+    func connect(_ peripheral: CBPeripheralType, options: [String: Any]?)
+    func scanForPeripherals(withServices serviceUUIDs: [CBUUID]?, options: [String: Any]?)
+    func stopScan()
+    func cancelPeripheralConnection(_ peripheral: CBPeripheralType)
+    func retrieveConnectedPeripherals(withServices serviceUUIDs: [CBUUID]) -> [CBPeripheralType]
+    func retrievePeripherals(withIdentifiers identifiers: [UUID]) -> [CBPeripheralType]
+}
+
+extension CBCentralManager: CBCentralManagerType {
+    var bluetoothState: BluetoothState {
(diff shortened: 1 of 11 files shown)


Source/CBPeripheralType.swift
```diff
@@ -0,0 +1,59 @@
...
+/// and CBPeripheral implements that protocol.
+/// It is needed for creating mocks of CBPeripheral in tests.
+
+protocol CBPeripheralType: class {
+    var delegate: CBPeripheralDelegate? { get set }
+    var state: CBPeripheralState { get }
+    var name: String? { get }
+    var services: [CBService]? { get }
+    var identifier: UUID { get }
+    var canSendWriteWithoutResponse: Bool { get }
```
dariuszseweryn: Isn't this property set per characteristic?

Source/CBPeripheralType.swift
```diff
@@ -0,0 +1,59 @@
...
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+// SOFTWARE.
+
+import CoreBluetooth
+
+/// CBPeripheralType is protocol that has exact same methods and variables as CBPeripheral
+/// and CBPeripheral implements that protocol.
+/// It is needed for creating mocks of CBPeripheral in tests.
+
+protocol CBPeripheralType: class {
```
dariuszseweryn: Is this protocol visible to the end user?

Source/CBCentralManagerType.swift
```diff
@@ -0,0 +1,65 @@
...
+import Foundation
+import CoreBluetooth
+
+/// CBCentralManagerType is protocol that has exact same methods and variables as CBCentralManager
+/// and CBCentralManager implements that protocol.
+/// The only difference here is bluetoothState - it is because in iOS 10 there was type change of
+/// state property and we can not create 2 variables with different type and same name.
+/// It is needed for creating mocks of CBCentralManager in tests.
+
+protocol CBCentralManagerType: class {
```
dariuszseweryn: Is this protocol visible to the end user?

paweljaneczek: no, only `public` is visible for end user

Source/CBPeripheralType.swift
paweljaneczek: same here

Source/CBPeripheralType.swift
paweljaneczek: it is only readable property, the same as in `CBPeripheral` class

Cierpliwy:
We had internal talk about this but I want to write about my concerns here as well.

First of all this approach is better then previous one, because it fixes problematic parts of a code related to RxCB wrappers. However in my opinion it is still not a perfect architecture, because there are visible chunks of code which are formed in a way to support tests (which end user don't care about):

- Force casting to CBPeripheral is a code smell:
```
/// Returns wrapped CBCentralManager instance
public func getCentralManager() -> CBCentralManager {
    return centralManager as! CBCentralManager
}
```
- Still we need to check implementation for each direct CB* function call with `as?`:
```
extension CBCentralManager: CBCentralManagerType {
    var bluetoothState: BluetoothState {
        return BluetoothState(rawValue: state.rawValue) ?? .unknown
    }
    func connect(_ peripheral: CBPeripheralType, options: [String: Any]?) {
        if let realPeripheral = peripheral as? CBPeripheral {
        ...
```
In a way it is the same story as with old `RxCB`  wrappers which were casting `Rx*Type` to `RxCB*`. There is unnecessary performance hit every call.
- Everything is behind protocol which could add another unnecessary performance hit.

I would encourage you to implement testing support with code generation, which will solve all above issues:

1. Write end user code without any protocols and by using pure CB* classes.
2. Add run-script which will generate implementation for testing:
    * Generate Fake* classes by scanning CB* classes and adding all required properties, functions with TestObservables.
    * Copy contents of CentralManager/Peripheral/Characteristic... classes to TestCentralManager/TestPeripheral/TestCharacteristic swift files.
    * Use simple regexp to change all CB* classes to Test* classes in copied files.
    * In XCode testing module use only Test* and Fake* classes.

Apart from that there are few things to keep in mind:

- We need to make sure that by default CBCentralManager delegate is setup in constructor to not miss any events.
- We need to make sure that there is no delegate stealing in Peripherals.
- We need to make sure that there are no memory leaks.

PR #202, opened 2018-01-12
Improvements to testing infrastructure

paweljaneczek:
- Added peripheral delegate provider to prevent delegate stealing (#163)
- Fixed not passing central manager delegate in constructor
- Code generation improvements - now delegate wrappers are also mocks, not testable classes


Source/CBPeripheral+Uuid.swift
```diff
@@ -0,0 +1,31 @@
...
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+// SOFTWARE.
+
+import Foundation
+import CoreBluetooth
+
+extension CBPeripheral {
+    /// There is still no identifier property for macOS, that's why we need to retrieve it by value method
+    var uuidIdentifier: UUID? {
+        return value(forKey: "identifier") as? NSUUID as UUID?
```
bartoszstelmaszuk: could we write here `value(forKey: "identifier") as! NSUUID as UUID` so we don't need to force cast `uuidIdentifier` every time

Source/BluetoothManager.swift
```diff
@@ -65,39 +67,39 @@ public class BluetoothManager {
     init(
```
Cierpliwy: update docs.

Tests/Autogenerated/_BluetoothManager.generated.swift
```diff
@@ -66,37 +68,30 @@ class _BluetoothManager {
     init(
```
Cierpliwy: update docs

Issue #167, opened 2017-10-23
[5.0] `monitorDisconnection` set of methods should emit `.next` along with `.error` in Observable stream


Kacper20:
Currently method that allows for monitoring disconnection of Peripheral is emitting `.next` on Observable stream, when proper delegate method fires (`centralManager(_:didDisconnectPeripheral:error)`). Semantically whenever other `BluetoothError` instances like are emitted (like `CBCentralManagerState` becomes .unauthorized) — the peripheral is also becoming disconnected. We would like in this case emit both `.next` and `.error` on the stream.

# Design
Inside the implementation for the monitor disconnection we should check for proper errors and map them into sequence of element and error. 

# How it affect current users
None affection

paweljaneczek:
Done in #206

PR #206, opened 2018-01-19
Simpler scanning and improved monitorDisconnection

paweljaneczek:
This PR is fixing 2 issues: #186 and #167 

As of #186, now if users starts new scanning when one is already ongoing it will return Error.

### DIFF
--- a/Source/BluetoothError.swift
+++ b/Source/BluetoothError.swift
@@ -29,6 +29,8 @@ public enum BluetoothError: Error {
     /// To mitigate it dispose all of your subscriptions before deinitializing
     /// object that created Observables that subscriptions are made to.
     case destroyed
+    // Emitted when `BluetoothManager.scanForPeripherals` called and there is already ongoing scan
+    case scanInProgress
     // States
     case bluetoothUnsupported
     case bluetoothUnauthorized
@@ -64,6 +66,11 @@ extension BluetoothError: CustomStringConvertible {
             return """
             The object that is the source of this Observable was destroyed.
             It's programmer's error, please check documentation of error for more details
+            """
+        case .scanInProgress:
+            return """
+            Tried to scan for peripheral when there is already ongoing scan.
+            You can have only 1 ongoing scanning, please check documentation of BluetoothManager for more details
             """
         case .bluetoothUnsupported:
             return "Bluetooth is unsupported"
@@ -134,6 +141,8 @@ extension BluetoothError: Equatable {}
 
 public func == (lhs: BluetoothError, rhs: BluetoothError) -> Bool {
     switch (lhs, rhs) {
+    case (.scanInProgress, .scanInProgress): return true
+    // States
     case (.bluetoothUnsupported, .bluetoothUnsupported): return true
     case (.bluetoothUnauthorized, .bluetoothUnauthorized): return true
     case (.bluetoothPoweredOff, .bluetoothPoweredOff): return true
--- a/Source/BluetoothManager.swift
+++ b/Source/BluetoothManager.swift
@@ -49,32 +49,26 @@ public class BluetoothManager {
 
     private let delegateWrapper: CBCentralManagerDelegateWrapper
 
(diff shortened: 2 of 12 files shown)


Source/BluetoothError.swift
```diff
@@ -134,6 +141,8 @@ extension BluetoothError: Equatable {}
 
 public func == (lhs: BluetoothError, rhs: BluetoothError) -> Bool {
     switch (lhs, rhs) {
+    case (.ongoingScan, .ongoingScan): return true
```
Cierpliwy: I'm not sure about name of this case. Maybe `scanInProgress` might be better?

Source/BluetoothManager.swift
```diff
@@ -378,12 +344,19 @@ public class BluetoothManager {
...
-    /// if it wasn't the `cancelConnection` call
+    /// if it wasn't the `cancelConnection` call. 
     public func monitorDisconnection(for peripheral: Peripheral) -> Observable<(Peripheral, DisconnectionReason?)> {
         let observable = delegateWrapper.didDisconnectPeripheral
           .filter { $0.0 == peripheral.peripheral }
           .map { (_, error) -> (Peripheral, DisconnectionReason?) in (peripheral, error) }
         return ensure(.poweredOn, observable: observable)
+                .catchError { error in
+                    if error is BluetoothError {
+                        return .concat(.just((peripheral, error)), .error(error))
```
Cierpliwy: Do we really need to emit .error in the end?

Tests/Autogenerated/Mock.generated.swift
```diff
@@ -21,8 +21,12 @@ class CBCentralManagerMock: NSObject {
 
     var retrievePeripheralsParams: [([UUID])] = []
     var retrievePeripheralsReturns: [[CBPeripheralMock]] = []
+    var retrievePeripheralsReturn: [CBPeripheralMock]?
     func retrievePeripherals(withIdentifiers identifiers: [UUID]) -> [CBPeripheralMock] {
         retrievePeripheralsParams.append((identifiers))
+        if let retrievePeripheralsReturn = retrievePeripheralsReturn {
```
Cierpliwy: Why is it necessary? If you want default return value, maybe we can use it after all return values from array are used?

Source/BluetoothManager.swift
```diff
@@ -128,70 +120,44 @@ public class BluetoothManager {
...
                         }
+                        .subscribe(observer)
+
+                strongSelf.centralManager.scanForPeripherals(withServices: serviceUUIDs, options: options)
+
+                return Disposables.create {
+                    guard let strongSelf = self else { return }
+                    // When disposed, stop scan and dispose scanning
+                    strongSelf.centralManager.stopScan()
+                    do { strongSelf.lock.lock(); defer { strongSelf.lock.unlock() }
```
Cierpliwy: We need to talk more about locking and thread safety of this library.

Tests/Autogenerated/Mock.generated.swift
paweljaneczek: :+1:

Source/BluetoothManager.swift
paweljaneczek: Don't we? I think sending error event when bluetooth state changed from .poweredOn is a must here.

Source/BluetoothManager.swift
paweljaneczek: in this issue or later ? :)

Source/BluetoothError.swift
paweljaneczek: ok, will do 👍

Source/BluetoothManager.swift
Cierpliwy: Yes, we need to talk about it later, before creating task for it. For now it can stay as it is.

Source/BluetoothManager.swift
Cierpliwy: For single peripheral is probably makes sense.

Issue #208, opened 2018-01-25
[5.0] Change `BluetoothManager` to `CentralManager`


paweljaneczek:
Since we are doing changes that will change current library API we could also change main class name. This will be heplfull in future when we will introduce `CBPeripheralManager` support

# Affection to current users
Rename of main user facing object is a source breaking change.

paweljaneczek:
Done in #209

PR #209, opened 2018-01-25
Refactored class and method names

paweljaneczek:
What has changed:

- changed `BluetoothManager` to `CentralManager`
- changed monitoring method to start with `observe` rather than `monitor` - it is more reactive and consistent with android library
- changed behaviour of CentralManager.observeState  and Peripheral.observeConnection to not start with current state
- added possibility to observing any events without specifing peripheral/characteristic/descriptor
- changed retrievePeripherals to synchronous methods
- grouped all method in code with marking

FIXED ISSUES: #208


Source/CentralManager+RestoredState.swift
```diff
@@ -0,0 +1,53 @@
...
+
+/// Closure that receives `RestoredState` as a parameter
+public typealias OnWillRestoreState = (RestoredState) -> Void
+
+extension CentralManager {
+
+    convenience init(queue: DispatchQueue = .main,
+                     options: [String: AnyObject]? = nil,
+                     onWillRestoreState: OnWillRestoreState? = nil) {
+        self.init(queue: queue, options: options)
```
Cierpliwy: Delegate will be created and bound to central in line 34.

Source/CentralManager+RestoredState.swift
```diff
@@ -0,0 +1,53 @@
...
+        if let onWillRestoreState = onWillRestoreState {
+            listenOnWillRestoreState(onWillRestoreState)
```
Cierpliwy: We start monitoring restoration state in line 36. There still could be a value emitted between line 34-36, which will not be caught.

paweljaneczek: Actually it will always work, cause `CBCentralManagerDelegateWrapper.willRestoreState` is of type  `ReplaySubject` with buffer == 1. It will buffer last value and pass it on subscription.

Issue #173, opened 2017-10-24
[5.0] Proposal: Making setNotification and connect APIs cleanup actions in subscriptions disposal


Kacper20:
This task is more of a discussion about the way we should re-imagine our APIs regarding setting notifications and making connections. It provides an overview on the current implementation and proposes some changes to this. If it’ll be accepted by the maintainers there should be a separate issue created with the detailed design. This two cases are considered together, because our goal is to have consistent APIs. *Remember*: This proposal It’s not necessarily the way to go — it’s only starting point for a discussion. The final solution could be this, sticking with the current approach or even mixing two of them. 

# Design
Currently in each case there are two separate methods: one that is changing state from A to B and the second one that is performing the opposite action. We would like to propose having abstract method:
```
func changeState() -> Observable<X>
```
This method would change state from A to B when subscription to resulting observable is made. The opposite action (transition from B to A) would be made on subscription disposal.
Let’s make this a bit more concrete:

## setNotification changes
The base for all of the notifications related methods is `setNotifyValue(_: Bool, for: Characteristic) -> Observable<Characteristic>`. We would like to propose changing it’s interface to: `setNotifications(on: Characteristic) -> Observable<Characteristic>`. This method would in the implementation set the notification and emit the next when the notification set was successful. This stream should never emit a Complete event. 
In Disposable for this method we would call  `setNotifyValue(false, for: CBCharacteristic)` on an underlying `CBPeripheral` instance. This will disable notifications on this characteristic. Additionally more work related to this observable being connectable will be performed in this piece of code.  Details can be found in `Common considerations for both changes` section.

## connect changes
Currently we have a set of two methods to perform connection related logic: `connect` and `cancelConnection`. To make it more reactive we could make it only one method —`connect`. This method would emit next when connection is made and will never emit a complete. Inside Disposed method we would perform disconnection.
Additionally more work related to this observable being connectable will be performed in this piece of code.  Details can be found in `Common considerations for both changes` section.

## Common considerations for both changes
Because of the nature of those Observable and the possibility that they could be subscribed to multiple times at once we should make them two Connectable Observables. They should be refcounted, so that cleanup is performed only after the last subscription is disposed. In order to implement this we should be storing Observables in some storages internal to BluetoothManager and Peripheral. 

# Affection on the user
It’s source breaking change. If there’ll be a decision to implement this it should be done in 5.0 version.

paweljaneczek:
Added in #212 and #215

PR #212, opened 2018-01-31
Changed connecting logic to handle auto disconnecting on dispose

paweljaneczek:
Part of #173.

`CentralManager.connect` method has changed to `CentralManager.establishConnection`. Now it is a more complex method that handles:
- auto disconnecting on dispose
- waiting for disconnecting
- throwing errors when there is already connected/connecting device.

All the changes are tested and documented

### DIFF
--- a/Source/Boxes.swift
+++ b/Source/Boxes.swift
@@ -35,3 +35,47 @@ extension WeakBox {
         return "WeakBox(\(String(describing: value)))"
     }
 }
+
+/**
+ `ThreadSafeBox` is a helper class that allows use of resource (value) in a thread safe manner.
+ All read and write calls are wrapped in concurrent `DispatchQueue` which protects writing to
+ resource from more than 1 thread at a time.
+ */
+class ThreadSafeBox<T>: CustomDebugStringConvertible {
+    private let queue = DispatchQueue(label: "com.polidea.RxBluetoothKit.ThreadSafeBox", attributes: .concurrent)
+    private var value: T
+    init(value: T) {
+        self.value = value
+    }
+
+    func read<Result: Any>(_ block: (T) -> Result) -> Result {
+        var result: Result! = nil
+        queue.sync {
+            result = block(value)
+        }
+        return result
+    }
+
+    func write(_ block: @escaping (inout T) -> Void) {
+        queue.async(flags: .barrier) {
+            block(&self.value)
+        }
+    }
+
+    func compareAndSet(compare: (T) -> Bool, set: @escaping (inout T) -> Void) -> Bool {
+        var result: Bool = false
+        queue.sync {
+            result = compare(value)
+            if result {
+                write(set)
+            }
+        }
+        return result
+    }
+}
+
+extension ThreadSafeBox {
+    var debugDescription: String {
+        return "ThreadSafeBox(\(String(describing: value)))"
+    }
+}
--- a/Source/Boxes.swift
+++ b/Source/Boxes.swift
@@ -42,7 +42,7 @@ extension WeakBox {
  resource from more than 1 thread at a time.
  */
 class ThreadSafeBox<T>: CustomDebugStringConvertible {
-    private let queue = DispatchQueue(label: "com.polidea.RxBluetoothKit.ThreadSafeBox", attributes: .concurrent)
+    private let queue = DispatchQueue(label: "com.polidea.RxBluetoothKit.ThreadSafeBox")
     private var value: T
     init(value: T) {
         self.value = value
@@ -57,7 +57,7 @@ class ThreadSafeBox<T>: CustomDebugStringConvertible {
     }
 
     func write(_ block: @escaping (inout T) -> Void) {
-        queue.async(flags: .barrier) {
+        queue.async {
             block(&self.value)
         }
     }
@@ -67,7 +67,7 @@ class ThreadSafeBox<T>: CustomDebugStringConvertible {
         queue.sync {
             result = compare(value)
             if result {
-                write(set)
+                set(&self.value)
             }
         }
         return result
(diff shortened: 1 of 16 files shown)


Source/Boxes.swift
```diff
@@ -35,3 +35,47 @@ extension WeakBox {
...
     }
 }
+
+/**
+ `ThreadSafeBox` is a helper class that allows use of resource (value) in a thread safe manner.
+ All read and write calls are wrapped in concurrent `DispatchQueue` which protects writing to
+ resource from more than 1 thread at a time.
+ */
+class ThreadSafeBox<T>: CustomDebugStringConvertible {
+    private let queue = DispatchQueue(label: "com.polidea.RxBluetoothKit.ThreadSafeBox", attributes: .concurrent)
```
Cierpliwy: Concurrent queue won't protect you from writing to resource from more then 1 thread.

Cierpliwy: You should switch to serial queue to achieve that.

Source/Boxes.swift
```diff
@@ -35,3 +35,47 @@ extension WeakBox {
...
+            block(&self.value)
+        }
+    }
+
+    func compareAndSet(compare: (T) -> Bool, set: @escaping (inout T) -> Void) -> Bool {
+        var result: Bool = false
+        queue.sync {
+            result = compare(value)
+            if result {
+                write(set)
```
Cierpliwy: no need to queue write inside sync queue.

Source/Connector.swift
```diff
@@ -0,0 +1,167 @@
...
+import RxSwift
+import CoreBluetooth
+
+/**
+ `Connector` is a class that is responsible for establishing connection with peripherals.
+ */
+class Connector {
+    let centralManager: CBCentralManager
+    let delegateWrapper: CBCentralManagerDelegateWrapper
+    let connectingBox: ThreadSafeBox<[UUID]> = ThreadSafeBox(value: [])
```
Cierpliwy: This should be a set instead of an array.

Source/Connector.swift
```diff
@@ -0,0 +1,167 @@
...
+    let disconnectingBox: ThreadSafeBox<[UUID]> = ThreadSafeBox(value: [])
```
Cierpliwy: This should be a set instead of an array.

Source/Connector.swift
```diff
@@ -0,0 +1,167 @@
...
+                observer.onError(BluetoothError.destroyed)
+                return Disposables.create()
+            }
+
+            guard !peripheral.isConnected else {
+                observer.onError(BluetoothError.peripheralAlreadyConnected(peripheral))
+                return Disposables.create()
+            }
+
+            let connectingStarted = strongSelf.connectingBox.compareAndSet(
```
Cierpliwy: I don't like that there is no global connection resource which is held uniquely. Current flow in theory may allow 2 simultaneously connected users. Let's assume that thread 1 and 2 are at line 96 and device is disconnected and not connecting to any other device. Thread 1 executes line 98 setting connecting state and after a while it goes to 138 by setting connecting to false after successful connection. Then thread 2 starts executing line 98 and is able to call `connect` even though thread 1 already owns a connection.

Cierpliwy: To clarify semantics we want to achieve there are three states which are important:

1) Device is disconnected. Any thread is allowed to take ownership of a resource.
2) Device is connecting/connected. One thread (winner) owns a resource, other threads reject with an error.
3) Device is disconnecting. Only owner from step 1 can transition to this state. Other threads now are allowed to wait.

paweljaneczek:
@MarvinNazari We think that it is better approach from developer perspective. Doing immediately `complete` without sending any `next` or `error` will cause developer not knowing what just happened.

Source/Connector.swift
```diff
@@ -90,17 +90,13 @@ class Connector {
...
-                compare: { !$0.contains(peripheral.identifier) },
-                set: { $0.append(peripheral.identifier) }
+            let connectionBlocked = strongSelf.connectingBox.compareAndSet(
+                compare: { !peripheral.isConnected && !$0.contains(peripheral.identifier) },
+                set: { $0.insert(peripheral.identifier) }
             )
-            guard connectingStarted else {
-                observer.onError(BluetoothError.peripheralIsConnecting(peripheral))
+            
+            guard connectionBlocked else {
```
paweljaneczek: what you think about changing name `connectionBlocked` to something like `connectingStarted`. For me it would be more readable

Tests/ThreadSafeBoxTest.swift
```diff
@@ -38,12 +38,13 @@ class ThreadSafeBoxTest: XCTestCase {
...
         
         let expectation = XCTestExpectation(description: "Perform \(expectedIterations) iterations on concurrent threads")
+        let resultQueue = DispatchQueue(label: "com.polidea.RxBluetoothKit.ThreadSafeBoxTest")
         
         DispatchQueue.concurrentPerform(iterations: currentIteration) { index in
             let last = box.read { $0.last } ?? 0
             box.write { $0.append(last + 1) }
             
-            DispatchQueue.global().sync {
+            resultQueue.sync {
```
paweljaneczek: 👍

Source/CentralManager.swift
```diff
@@ -185,81 +192,19 @@ public class CentralManager {
...
-    /// returned observable cancels connection attempt. By default observable is waiting infinitely for successful connection.
-    /// Additionally you can pass optional [dictionary](https://developer.apple.com/library/ios/documentation/CoreBluetooth/Reference/CBCentralManager_Class/#//apple_ref/doc/constant_group/Peripheral_Connection_Options)
+    /// Establishes connection with a given `Peripheral`.
+    /// When connection did succeded it sends event with `Peripheral` - from now on it is possible to call all other methods that require connection.
+    /// The connection is automatically disconnected when resulting Observable is unsubscribed.
+    /// On the other hand when the connection is interrupted or failed by the device or the system, the Observable will be unsubscribed as well
+    /// following `BluetoothError.peripheralConnectionFailed` or `BluetoothError.peripheralDisconnected` emission.
+    /// Additionally you can p ass optional [dictionary](https://developer.apple.com/library/ios/documentation/CoreBluetooth/Reference/CBCentralManager_Class/#//apple_ref/doc/constant_group/Peripheral_Connection_Options)
```
Cierpliwy: pass?

Issue #207, opened 2018-01-25
[5.0] Add unit tests to project


paweljaneczek:
With #168 we have introduced new testing architecture and deleted all existing unit tests.

We need to re-write all unit tests using this new architecture and clean XCTest framework.

# Classes that we need write tests for

- [ ] `BluetoothManager` (`CentralManager`)
- [ ] `Peripheral`, `Peripheral+Convienence`
- [ ] `Observable+Absorb`
- [ ] `UUIDIdentifiable`
- [ ] `ScanOperation`
- [ ] `ScanOperation`

In all other classes there is no logic so tests are optional.

# Affection to current users
This change doesn't affect current users.

paweljaneczek:
Added in #215

PR #215, opened 2018-03-05
New notifications observing logic, Peripheral provider and unit tests.

minixT:
Part of #173 and #207 

A few changes that update API:

- added `PeripheralProvider` - central place constructing and managing `Peripheral` objects
- added `CharacteristicNotificationManager` - which is used to observe characteristics value update and change notification flags on characteristics
- discovering services, includedServices and characteristics will emit `RxError.noElements` error, if not all objects with requested UUIDs were discovered

Added unit tests to the missing classes.


Source/CharacteristicNotificationManager.swift
```diff
@@ -0,0 +1,98 @@
...
+            }
+
+            let notifyValue = strongSelf.setNotifyValue(true, for: characteristic).asObservable()
+            let notificationObserable = strongSelf.createValueUpdateObservable(for: characteristic)
+            let observable = Observable.of(notifyValue, notificationObserable)
+                .merge()
+                .do(onDispose: { [weak self] in
+                    guard let strongSelf = self else { return }
+                    do { strongSelf.lock.lock(); defer { strongSelf.lock.unlock() }
+                        strongSelf.uuidToActiveObservableMap.removeValue(forKey: characteristic.uuid)
```
paweljaneczek: I am just thinking of situation when:
1. Thread 1 stops at line 53
2. Thread 2 locks at line 42 and taking observable on line 44-45
3. Thread 1 removing observable from map at line 55
In this screnario Thread 2 receives observable that is disposed. Let's talk about this when you will be at work

Source/CharacteristicNotificationManager.swift
```diff
@@ -0,0 +1,98 @@
...
+                    }
+                    _ = strongSelf.setNotifyValue(false, for: characteristic)
```
paweljaneczek: Not sure if we would not need to wait here for `CBPeripheralDelegate peripheral:didUpdateNotificationStateForCharacteristic:error:`
There could be scenario like:
1. Thread 1 stops on line 57 (`setNotifyValue` not called)
2. Thread 2 goes to line 62, subscibes for observable and calls `setNotifyValue(true)`
3. Thread 1 is doing `setNotifyValue(false)` on line 57
In result we have `setNotifyValue` turned to false and existing observable for notification change (which will never receive notification)

Simple solution would be to wrap this line in `do lock` from line 54. In that case it would be not possible to call `setNotifyValue`  with false after `setNotifyValue` with true. Question here is, how will CoreBluetooth behave when we will set 2x `setNotifyValue` one after another

Source/Peripheral+Convenience.swift
```diff
@@ -159,26 +159,18 @@ extension Peripheral {
...
-    /// This is **infinite** stream of values.
+    /**
+     Setup characteristic notification in order to receive callbacks when given characteristic has been changed.
+     Returned observable will emit `Characteristic` on every notification change.
+     It is possible to setup more observables for the same characteristic and the lifecycle of the notification will be shared among them.
+     
+     Notification is automaticaly unregistered once this observable is unsubscribed
+     
+     - parameter characteristic: `Characteristic` for notification setup.
+     - returns: `Observable` emitting `Peripheral` when the notification setup is complete.
```
paweljaneczek: same here for doc

Source/Peripheral.swift
```diff
@@ -330,51 +347,21 @@ public class Peripheral {
...
-            .merge()
+    /**
+     Setup characteristic notification in order to receive callbacks when given characteristic has been changed.
+     Returned observable will emit `Characteristic` on every notification change.
+     It is possible to setup more observables for the same characteristic and the lifecycle of the notification will be shared among them.
+     
+     Notification is automaticaly unregistered once this observable is unsubscribed
+     
+     - parameter characteristic: `Characteristic` for notification setup.
+     - returns: `Observable` emitting `Peripheral` when the notification setup is complete.
```
paweljaneczek: same here

Source/PeripheralProvider.swift
```diff
@@ -0,0 +1,100 @@
...
+        } else {
+            return createAndAddToBox(cbPeripheral, manager: centralManager)
+        }
+    }
+
+    fileprivate func createAndAddToBox(_ cbPeripheral: CBPeripheral, manager: CentralManager) -> Peripheral {
+        peripheralsBox.compareAndSet(
+            compare: { peripherals in
+                return !peripherals.contains(where: { $0.peripheral == cbPeripheral })
+        },
```
paweljaneczek: add tab here

paweljaneczek:
@rohitjb agree that we could split doing tests to other PR, but due to big delays in reviewing we wanted to make it in one shoot. It won't happen again in future ;)

Source/CharacteristicNotificationManager.swift
```diff
@@ -75,24 +82,14 @@ class CharacteristicNotificationManager {
...
-                    throw BluetoothError.characteristicNotifyChangeFailed(characteristic, error)
-                }
-                return characteristic
-        }
-        let setObservable: Observable<Characteristic> = .deferred { [weak self] in
+    private func setNotifyValue(_ enabled: Bool, for characteristic: Characteristic) -> Completable {
+        return .deferred { [weak self] in
+            if enabled {
+                let counter = self?.uuidToActiveObservablesCountMap[characteristic.uuid] ?? 0
+                self?.uuidToActiveObservablesCountMap[characteristic.uuid] = counter + 1
```
paweljaneczek: I think that it should be also guarded by NSLock. 
Imagine scenario:
1. thread 1 is on line 89, after getting counter value (value==1)
2. thread 2 goes to line 58 and decrease counter (value==0)
3. thread 1 changes coun map on line 89 (value==2)
Finally we have count==2 and only 1 observable.

PR #218, opened 2018-03-07
Documentation update

minixT:
Updated Readme file and code comments to match api changes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,7 +54,7 @@ Versions >= 4.0 of the library integrate with the Swift Package Manager. In orde
 Library is built on top of Apple's CoreBluetooth.
 It has multiple components, that should be familiar to you:
 
-- BluetoothManager
+- CentralManager
 - ScannedPeripheral
 - Peripheral
 - Service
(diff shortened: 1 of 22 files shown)


README.md
```diff
@@ -106,26 +106,28 @@ As you can see: thanks to all available RxSwift operators, in a simple way you m
...
+This observable will emit next event with new value of BluetoothManager state every time state changes.
+You could easily chain it with operation you want to perform after changing to proper state. To get current bluetooth state, use `CentralManager.state`
+property. Let's see how it looks with scanning:
 ```swift
-manager.rx_state
+manager.observeState
+	.startWith(manager.state)
 	.filter { $0 == .poweredOn }
 	.timeout(3.0, scheduler)
 	.take(1)
```
paweljaneczek: I think this `take(1)` is not needed. There won't be more than 1 event due to `ensure(.poweredOn` in `CentralManager`

paweljaneczek: hmm, question is, it is possible that `CoreBluetooth` will send `poweredOn` twice?

Source/Peripheral.swift
```diff
@@ -73,13 +76,13 @@ public class Peripheral {
         return peripheral.state == .connected
     }
 
-    ///  Current state of `Peripheral` instance described by [CBPeripheralState](https://developer.apple.com/library/ios/documentation/CoreBluetooth/Reference/CBPeripheral_Class/#//apple_ref/c/tdef/CBPeripheralState).
+    ///  Current state of `Peripheral` instance described by [CBPeripheralState](https://developer.apple.com/documentation/corebluetooth/cbperipheralstate).
```
paweljaneczek: 👍

README.md
minixT: hmm, I think CoreBluetooth sends this state each time user re-enable bluetooth in his device

paweljaneczek: yeah, but when it sends `poweredOn`, `poweredOff`, `poweredOn` then subscription is disposed due to `poweredOff` so `poweredOn` won't be called.
Problem might be when we will receive `poweredOn`, `poweredOn`, than without `take(1)` we will 2x start scanning.

PR #230, opened 2018-03-15
Characteristics state changes can now be observed

kkosovsky:
Added public method observeCharacteristicStateChanged(:_)
to Peripheral class, which maps events coming from
delegateWrapper.peripheralDidUpdateNotificationStateForCharacteristic
to determine when given Characteristic has changed it's state
(e.g isNotyfing). CBPeripheralDelegateWrapper field is private
in the Characteristic class, hence the need for providing public
access to it's functionality through Peripheral.

Also .characteristicStateChangedFailed was added to BluetoothError.

### DIFF
--- a/Source/BluetoothError.swift
+++ b/Source/BluetoothError.swift
@@ -28,6 +28,7 @@ public enum BluetoothError: Error {
     case characteristicWriteFailed(Characteristic, Error?)
     case characteristicReadFailed(Characteristic, Error?)
     case characteristicNotifyChangeFailed(Characteristic, Error?)
+    case characteristicStateChangedFailed(Characteristic, Error?)
     // Descriptors
     case descriptorsDiscoveryFailed(Characteristic, Error?)
     case descriptorWriteFailed(Descriptor, Error?)
@@ -61,7 +62,7 @@ extension BluetoothError: CustomStringConvertible {
             return "Bluetooth is in unknown state"
         case .bluetoothResetting:
             return "Bluetooth is resetting"
-            // Peripheral
+        // Peripheral
         case .peripheralIsConnectingOrAlreadyConnected:
             return """
             Peripheral is already connected or is in connecting state.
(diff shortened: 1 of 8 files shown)


Source/BluetoothError.swift
```diff
@@ -28,6 +28,7 @@ public enum BluetoothError: Error {
     case characteristicWriteFailed(Characteristic, Error?)
     case characteristicReadFailed(Characteristic, Error?)
     case characteristicNotifyChangeFailed(Characteristic, Error?)
+    case characteristicStateChangedFailed(Characteristic, Error?)
```
minixT: I'd prefer adding to this error name an information that it is related to notifications. We will get this error after calling `setNotifyValue` method on `Peripheral` (https://developer.apple.com/documentation/corebluetooth/cbperipheral/1518949-setnotifyvalue), so maybe something like `characteristicSetNotifyFailed` would be better?

Source/BluetoothError.swift
```diff
@@ -88,7 +89,9 @@ extension BluetoothError: CustomStringConvertible {
             return "Characteristic read error has occured: \(err?.localizedDescription ?? "-")"
         case let .characteristicNotifyChangeFailed(_, err):
             return "Characteristic notify change error has occured: \(err?.localizedDescription ?? "-")"
-            // Descriptors
+        case let .characteristicStateChangedFailed(_, err):
+            return "Characteristic state change error has occured: \(err?.localizedDescription ?? "-")"
```
minixT: Here we need add an information that this error occured during setting notify value.

Source/Peripheral.swift
```diff
@@ -347,6 +347,19 @@ public class Peripheral {
         return ensureValidPeripheralState(for: observable)
     }
 
+    public func observeCharacteristicStateChanged(for characteristic: Characteristic) -> Observable<Characteristic> {
```
minixT: You didn't add comment with documentation to this method

minixT: I think name of this method is not good. We don't need characteristic part - we know this from parameters. This is not also changing characteristic state. In [documentation](https://developer.apple.com/documentation/corebluetooth/cbperipheral/1518949-setnotifyvalue) there is not information about characteristic state, it is notifying or not notifying, so maybe we can change this method name to `observeNotifyValue`

Issue #238, opened 2018-03-23
[5.0] Unable to get an Observable<Peripheral> from already connected peripheral provided by the system


Drusy:
In v4 it was possible to call `connect` on an already connected peripheral without error, so we could have an observable on every single Peripheral, wherever it comes (scan / restoration / retrieving)

Now calling `establishConnection` on an already connected peripheral is forbidden. So there is no way to get an `Observable<Peripheral>` from peripherals that comes from state restoration as it will provider a `[Peripheral]`.

Same problem for `retrieveConnectedPeripherals(_) -> [Peripheral]`. So basically there is no way to disconnect those peripherals using RxBluetoothKit, the only way I see is to use the CBCentralManager directly and call `manager.centralManager.cancelPeripheralConnection(connectedPeripheral.peripheral)`

Discussed here https://gitter.im/RxBLELibraries/RxBluetoothKit with @pawelByszewski:

> We are thinking of making possible to call establishConnection on already connected peripheral. It would work the same as reading characteristic value, so we would do ref counting on subscriptions and when ref count goes to 0 then dispose & disconnect happens.
> Those solutions would not prevent user from their mistakes (need to dispose every subscription to disconnect device), but would surely give them more freedom and would fix problems that we currently have.
> What you think about it? We really want to make library that is easy to use and gives all the possibility as CBCentralManager gives.

> We will need to discuss that problem internally at Polidea and will give some final proposition next week

RamblinWreck77:
In my project that's exactly what I had to do, drill down to central manager and disconnect from the peripheral if I happened to already be connected.  Not a deal breaker, but definitely not ideal... and it kind of defeats the purpose of 5.0's auto-disconnect feature. Since this was necessary in 4.x anyways I was already used to doing this. 

I'm interested to see how they evolve this over time.

DevAndArtist:
@minixT I wonder if we can fix the `establishConnection` function by doing something like this (not tested code):

```swift
fileprivate func createConnectionObservable(
    for peripheral: Peripheral,
    options: [String: Any]? = nil
) -> Observable<Peripheral> {
  return Observable.create { [weak self] observer in
    guard let strongSelf = self else {
      observer.onError(BluetoothError.destroyed)
      return Disposables.create()
    }

    let connectingStarted = strongSelf.connectingBox.compareAndSet(
      compare: { !$0.contains(peripheral.identifier) },
      set: { $0.insert(peripheral.identifier) }
    )

[...]

PR #242, opened 2018-04-04
Enable disconnection for peripheral connected outside library (#238)

minixT:
In version 5.0 it was not possible to disconnect from peripheral which was connected outside library or  was provided by the system (e.g. it comes from state restoration). This pull request changes this situation. 

It will still be impossible to call `establishConnection` twice until the first connection is active.


paweljaneczek:
@minixT please rebase on newest master, it can be then marged

PR #243, opened 2018-04-04
Added availabilty check in writeValue(:_) method in Peripheral.swift

kkosovsky:
CBCharacteristicWriteType.withoutResponse was introduced in iOS 11.
Prior to this commit, there was no availability check performed,
which was causing a crash on devices with system version lower
than iOS 11. Now if the device's system is lower, we use
.withResponse case.


paweljaneczek:
@kkosovsky please rebase to current master before merging

Issue #235, opened 2018-03-22
[5.0] Reading and observing values using CharacteristicIdentifier onError before didDiscoverCharacteristics


Drusy:
Since v5.0, chaining `establishConnection` and `observeValueUpdateAndSetNotification`/`readValue` produces the RxSwift error `Sequence doesn't contain any elements`.

### Conditions

- Happens when chaining instantly the read or/and notify for more than once characteristic.
- Always succeed after few retry (1~3 x2s)
- Fails only after `establishConnection`, always works if the peripheral comes from `retrieveConnectedPeripherals` (services & characteristics already in memory ?)

### Related code:

```swift
enum DeviceCharacteristic: String, CharacteristicIdentifier {
    
    // Monitoring
    case temperature = "1b0d1202-a720-f7e9-46b6-31b601c4fca1" // °C
    case battery = "1b0d1203-a720-f7e9-46b6-31b601c4fca1" // mV

    var uuid: CBUUID {
        return CBUUID(string: self.rawValue)
    }
    
    //Service to which characteristic belongs
    var service: ServiceIdentifier {
        switch self {
        case .temperature, .battery:
            return DeviceService.monitoring
    }
}

enum DeviceService: String, ServiceIdentifier {
    case monitoring = "1b0d1200-a720-f7e9-46b6-31b601c4fca1"

    var uuid: CBUUID {
        return CBUUID(string: self.rawValue)
    }
}
```

```swift
self.manager.observeState()
    .startWith(manager.state)
    .filter { $0 == .poweredOn }
    .take(1)
    .flatMap {  peripheral.establishConnection() }
    .subscribe(
        onNext: { connectedPeripheral in
            self.onDeviceConnected(for: connectedPeripheral)
    },
        onError: { error in
            print("❌ Could not connect \(peripheral.name ?? ""): \(error)")
    }).disposed(by: self.currentConnectionDisposeBag)
```

``` swift
func onDeviceConnected(for: Peripheral) {
    print("🔵 \(peripheral.name ?? "") is now connected")

    // Read
    PrimitiveSequence.zip(
        peripheral.readValue(for: DeviceCharacteristic.battery).do(onSuccess: { _ in
            print("✅ Read battery")
        }),
        peripheral.readValue(for: DeviceCharacteristic.temperature).do(onSuccess: { _ in
            print("✅ Read temperature")
        }))
        .retryWhen { error in
            error.do(onNext: { error in
[...]

PR #244, opened 2018-04-06
Added "requests queues" to the discovery operations.

minixT:
This commit resolves problems with discovery operations when app wants to discover services, includedServices or characteristics. For example when app calls `discoverServices` two times in a row with different uuids and second call happen before getting response for the first call there was a problem in current version. In this situation observer that subscribe on the second request will get an error instead waiting for response for its requests.

This pull request introduce a new logic to the discovery operations. Now observables will not return any result until all requested components are discovered or all requests returns a result.

It resolves also #235

### DIFF
--- a/Source/Peripheral.swift
+++ b/Source/Peripheral.swift
@@ -3,6 +3,7 @@ import RxSwift
 import CoreBluetooth
 
 // swiftlint:disable line_length
+// swiftlint:disable type_body_length
 
 /// Peripheral is a class implementing ReactiveX API which wraps all Core Bluetooth functions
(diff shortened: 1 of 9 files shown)


Source/Peripheral.swift
```diff
@@ -19,6 +20,17 @@ public class Peripheral {
 
     let delegateWrapper: CBPeripheralDelegateWrapper
 
+    private let remainingServicesDiscoveryRequest = ThreadSafeBox<Int>(value: 0)
+    private let peripheralDidDiscoverServices = PublishSubject<([CBService]?, Error?)>()
```
paweljaneczek: what is this for? I don't see any usage of this property

Issue #245, opened 2018-04-06
[5.0] Error case `peripheralIsConnectingOrAlreadyConnected` is misleading


DevAndArtist:
I reviewed the changes made in https://github.com/Polidea/RxBluetoothKit/pull/242 and from the standpoint of the new implementation the error case becomes misleading because it will only fail if one observable was previously created.

I'd propose to rename the case to something like `isAlreadyObservingConnection`.

paweljaneczek:
@DevAndArtist Hey, good point. Could you look at https://github.com/Polidea/RxBluetoothKit/pull/249 and check if it looks ok?

paweljaneczek:
Fixed in #249

PR #249, opened 2018-04-18
Changed peripheral already connected error

paweljaneczek:
- changed Error.peripheralIsConnectingOrAlreadyConnected to  Error.pe…ripheralIsAlreadyObservingConnection so it better describes what happened
- updated sourcery (newest swift)
- added IDEWorskapceChecks.plist that is generated on xCode 9.3

FIXED ISSUES: #245


DevAndArtist:
@pawelByszewski the new error case looks good to me. 👍

Issue #225, opened 2018-03-14
Update API documentation with a list of errors


DevAndArtist:
Please update documentation for all properties and functions that return an `Observable` or a different RxSwift type with a list of all possible errors that particular sequence can throw. It is literally very painful to recursively dig through the entire implementation to find all possible errors that can happen on each sequence.

paweljaneczek:
@DevAndArtist I've added it in  #250. Please take a look if you will have time

paweljaneczek:
Done in #250

PR #250, opened 2018-04-19
Update API documentation with a list of errors

paweljaneczek:
Added info what error can be returned by observable to each method that returns observable.

FIXED ISSUES: #225


DevAndArtist:
@paweljaneczek I think you should also put the errors that will break an observable sequence which are delegated from a different observable we depend on. The thing is that not all observables do emit errors.

Some examples:

- On `CentralManager` not all observables emit errors. `observeState` won't emit any errors. That's why you should put the errors always in the comments above the type member that returns an observable instead above the type itself.

- `establishConnection` on a peripheral will also emit all the errors that the manager will delegate because it depends on the bluetooth state. Therefore those errors should also be part of that documentation since the API user does not have to read the implementation to find out that it will call the method on the manager with `self` and fail with the same errors as the other method.

I know this is tedious, but it has to be done, that's what I did for myself when building a wrapper API around this module and it took a good amount of my time (however I wasn't 100% sure that I spotted all of them for each observable that I used - that's why I opened the issue in the first place).

paweljaneczek:
@DevAndArtist hmm yes, you are right. I wanted to make it more cleaner (there are a lot of default error related to state changes) but with such approach it not gives all the info that we want to give to user. Will change that. Thanks!

paweljaneczek:
@DevAndArtist could you please check it again?

DevAndArtist:
@pawelByszewski that's what I've been expecting from the the beginning, thanks. 👍

PR #292, opened 2018-09-25
updated cartfile to point RxSwift to 4.3.1

raviranderia:


Cartfile
```diff
@@ -1 +1 @@
-github "ReactiveX/RxSwift" ~> 4.0
+github "ReactiveX/RxSwift" ~> 4.3.1
```
bartoszstelmaszuk: Basically `~> 4.0` will do almost the same as `~> 4.3.1`. It means "install RxSwift with version at least 4.0 but lower than 5.0". What it needs to be done is to run `Cartfile update` to get new version of RxSwift to library (see `Cartfile.resolved`)

raviranderia: I did run carthage, but it was still always getting 4.0 for some reason.

bartoszstelmaszuk: did you run firstly `carthage upgrade`?

raviranderia: yes

PR #320, opened 2019-03-18
remove `self` = self

bmbowdish:
`self`= self is a known compiler bug and it seems like you have fixed it or avoided it in the rest of the project. Found this in the Example App when using it to learn about RxBluetoothKit

### DIFF
--- a/ExampleApp/ExampleApp/Common/Services/RxBluetoothKitService.swift
+++ b/ExampleApp/ExampleApp/Common/Services/RxBluetoothKitService.swift
@@ -94,10 +94,10 @@ final class RxBluetoothKitService {
         .subscribeOn(MainScheduler.instance)
         .timeout(4.0, scheduler: scheduler)
         .flatMap { [weak self] _ -> Observable<ScannedPeripheral> in
-            guard let `self` = self else {
+            guard let strongSelf = self else {
                 return Observable.empty()
             }
-            return self.centralManager.scanForPeripherals(withServices: nil)
+            return strongSelf.centralManager.scanForPeripherals(withServices: nil)
         }.subscribe(onNext: { [weak self] scannedPeripheral in
                     self?.scanningSubject.onNext(Result.success(scannedPeripheral))
                 }, onError: { [weak self] error in


stefanomondino:
why not simply use `guard let self = self` instead of `strongSelf`? It's possible since swift 4.2

minixT:
Hi @bmbowdish! Thank you for your PR.

You are right, using \`self\` in the example app is inconsistency with the library code where we are using `strongSelf`. As @stefanomondino mentioned now we can use `let self = self` wherever we want to unwrap self. @bmbowdish do you want to change this in the library code to use `let self = self` in all places where it is possible?

bmbowdish:
Sure, almost done.

bmbowdish:
@minixT I accidentally used a different email address, but I added it to my GitHub account. It should be working.

Source/CharacteristicNotificationManager.swift
```diff
@@ -17,35 +17,35 @@ class CharacteristicNotificationManager {
...
                 .do(onSubscribed: { [weak self] in
-                    guard let strongSelf = self else { return }
-                    strongSelf.lock.lock(); defer { strongSelf.lock.unlock() }
-                    let counter = strongSelf.uuidToActiveObservablesCountMap[characteristic.uuid] ?? 0
-                    strongSelf.uuidToActiveObservablesCountMap[characteristic.uuid] = counter + 1
+                    guard let self = self else { return }
+                    self.lock.lock(); defer { self.lock.unlock() }
+                    let counter = self.uuidToActiveObservablesCountMap[characteristic.uuid] ?? 0
+                    self.uuidToActiveObservablesCountMap[characteristic.uuid] = counter + 1
                     self?.setNotifyValue(true, for: characteristic)
```
ptseng: `self` is now `self`

Tests/Autogenerated/_CharacteristicNotificationManager.generated.swift
```diff
@@ -18,35 +18,35 @@ class _CharacteristicNotificationManager {
...
                 .do(onSubscribed: { [weak self] in
-                    guard let strongSelf = self else { return }
-                    strongSelf.lock.lock(); defer { strongSelf.lock.unlock() }
-                    let counter = strongSelf.uuidToActiveObservablesCountMap[characteristic.uuid] ?? 0
-                    strongSelf.uuidToActiveObservablesCountMap[characteristic.uuid] = counter + 1
+                    guard let self = self else { return }
+                    self.lock.lock(); defer { self.lock.unlock() }
+                    let counter = self.uuidToActiveObservablesCountMap[characteristic.uuid] ?? 0
+                    self.uuidToActiveObservablesCountMap[characteristic.uuid] = counter + 1
                     self?.setNotifyValue(true, for: characteristic)
```
ohtwo: Should change like `self.setNotifyValue` instead optional self

Issue #319, opened 2019-03-14
API MISUSE after disposing observeValueUpdateAndSetNotification


David2XN:
**Describe the bug**

Following error occurs in Xcode Output after disposing disposable from observeValueUpdateAndSetNotification() when bluetooth is Off:
```
2019-03-14 15:56:18.536575+0100 test[286:5547] [CoreBluetooth] API MISUSE: <CBPeripheral: 0x282835040, identifier = FCADF30F-882E-3766-212A-81915268838D, name = pL*****, state = disconnected> can only accept commands while in the connected state
2019-03-14 15:56:18.536898+0100 test[286:5547] [CoreBluetooth] API MISUSE: <CBCentralManager: 0x280c407c0> can only accept this command while in the powered on state
```

**To Reproduce**
1. Connect to BLE device
2. Discover services and Characteristics
3. Call observeValueUpdateAndSetNotification(), subscribe to it and store its disposable (observeDisposable)
4. Turn Off Bluetooth in Control Center
5. observeDisposable?.dispose() = Dispose disposable from observeValueUpdateAndSetNotification() (performed in .observeConnection().filter{ !$0 })

**Expected behavior**
No Errors in XCode Output

**Environment:**
 - Device: iPhone SE
 - OS: 12.1.4 (build 16D57)
 - Library version: 5.1.4
 - Swift version: 4.2.1

minixT:
Fixed in version 5.2.0

PR #323, opened 2019-03-31
Fixed issue #319 causing "API MISUSE" warning from CoreBluetooth

wojciech-kulik:
Fixed issue #319 causing "API MISUSE" warning from CoreBluetooth


Source/CharacteristicNotificationManager.swift
```diff
@@ -63,6 +63,7 @@ class CharacteristicNotificationManager {
     }
 
     private func setNotifyValue(_ enabled: Bool, for characteristic: Characteristic) {
+        guard peripheral.state == .connected else { return }
```
minixT: Please add logs to inform when this condition is not passed.
```suggestion
        guard peripheral.state == .connected else {
                RxBluetoothKitLog.w("""
                        \(peripheral.logDescription) is not connected. Changing notification state for not connected peripheral is not possible.
                        """)
                return 
        }
```

Source/Connector.swift
```diff
@@ -97,7 +97,9 @@ class Connector {
                 let isConnected = strongSelf.connectedBox.read { $0.contains(peripheral.identifier) }
                 if isConnected {
                     strongSelf.disconnectingBox.write { $0.insert(peripheral.identifier) }
-                    strongSelf.centralManager.cancelPeripheralConnection(peripheral.peripheral)
+                    if strongSelf.centralManager.state == .poweredOn {
+                        strongSelf.centralManager.cancelPeripheralConnection(peripheral.peripheral)
```
minixT: You should call also this line `strongSelf.disconnectingBox.write { $0.insert(peripheral.identifier) }` only when bluetooth is enabled.

wojciech-kulik: Done

Source/CharacteristicNotificationManager.swift
wojciech-kulik: Done, I splitted log message into two lines, because suggested change caused swiftlint warning.

PR #337, opened 2019-05-17
Always connect to a peripheral before trying to communicate with it

emixb:
Follow Apple's recommendation and always connect to a peripheral before trying to communicate with it. This fixes a state restoration issue where the app relaunched in background couldn't communicate with known peripheral. The problem was that the reported state of the restored peripheral was connected and RxBluetoothKit would skip connecting to it. As Apple documents it everywhere, we should connect to a peripheral in all cases. "Even though the device is already connected to the system, you must still connect it locally to your app to begin exploring and interacting with it"


minixT:
Hi @emixb!

Thank you for your PR. Can you update tests in the project? One test does not pass after your changes.

emixb:
> One test does not pass after these changes. Please update project tests.

Hi @minixT, I updated the tests. Thanks!

PR #341, opened 2019-05-21
Add a customisation point for injecting a logger in RxBluetoothKit

asaltvld30:
Adds the capability to route the RxBluetoothKit logs in another app by exposing a customisation point.
Each custom logger must conform to **Logger** protocol.

The current functionality remains unchanged unless another logger is injected.


Source/RxBluetoothKitLogger.swift
```diff
@@ -0,0 +1,126 @@
...
+
+    func log(
+        _ message: @autoclosure () -> String,
+        level: RxBluetoothKitLog.LogLevel,
+        file: String,
+        function: String,
+        line: UInt
+    ) {
+        if currentLogLevel <= level {
+            print("\(tag(with: level)) \(message())")
```
minixT: Now when you add information about file, function and line, we can use this information in this standard logger and print it as well.

Source/Logging.swift
```diff
@@ -28,69 +26,65 @@ public class RxBluetoothKitLog: ReactiveCompatible {
...
-                               floor(time / 3600.0).truncatingRemainder(dividingBy: 24),
-                               floor(time / 60.0).truncatingRemainder(dividingBy: 60),
-                               floor(time).truncatingRemainder(dividingBy: 60),
-                               floor(time * 1000).truncatingRemainder(dividingBy: 1000))
-    }
-
-    fileprivate static func log(with logLevel: LogLevel, message: @autoclosure () -> String) {
-        if currentLogLevel <= logLevel {
-            let string = "\(tag(with: logLevel)) \(message())"
-            subject.onNext(string)
```
minixT: You removed possibility to use `log` observable to observe logs from the library. Please add this to the new `log` method.

Source/Logging.swift
```diff
@@ -28,69 +26,65 @@ public class RxBluetoothKitLog: ReactiveCompatible {
         case error = 4
     }
 
-    /// Set new log level.
-    /// - Parameter logLevel: New log level to be applied.
-    public static func setLogLevel(_ logLevel: LogLevel) {
```
minixT: `setLogLevel` and `getLogLevel` methods should be left in `RxBluetoothKitLog` class. These methods are used in some projects and removing them will cause these projects will not compile. My suggestion is to add `setLogLevel` and `getLogLevel` methods to `Logger` protocol what will enable to use them in this class.

Source/RxBluetoothKitLogger.swift
```diff
@@ -0,0 +1,126 @@
...
+        )
+    }
+}
+
+/// Simple logging interface.
+///
+/// An application that wants RxBluetoothKit to use its logging solution will
+/// need to provide a type that conforms to this signature and assign it to
+/// `RxBluetoothKitLogger.defaultLogger`.
+public protocol Logger {
```
minixT: Can you move `Logger` protocol to separate file?

minixT:
Can you also prepare some example how to use custom logger which I can add to the wiki page?

asaltvld30:
> Can you also prepare some example how to use custom logger which I can add to the wiki page?

Sure.
A custom logger can be created by simply implementing the **Logger** protocol which specifies  two logging methods and log level getters and setters.

For example, we can create a simple logger which counts the number of logs for each log level.

```swift
private class SimpleCountLogger: Logger {
    private var logCount: [UInt]

    private var currentLogLevel: RxBluetoothKitLog.LogLevel = .verbose

    init() {
        logCount = [UInt](repeating: 0, count: Int(UInt8.max))
    }

    public func getLogCount() -> [UInt] {
        return logCount
    }

    public func setLogLevel(_ logLevel: RxBluetoothKitLog.LogLevel) {
        self.currentLogLevel = logLevel
    }

    public func getLogLevel() -> RxBluetoothKitLog.LogLevel {
        return currentLogLevel
    }

    func log(
        _ message: @autoclosure () -> String,
        level: RxBluetoothKitLog.LogLevel,
        file: StaticString,
        function: StaticString,
        line: UInt
    ) {
        log(
            message(),
            level: level,
            file: String(describing: file),
            function: String(describing: function),
            line: line
        )
    }

    func log(
        _ message: @autoclosure () -> String,
        level: RxBluetoothKitLog.LogLevel,
        file: String,
        function: String,
        line: UInt
    ) {
        logCount[Int(level.rawValue)] += 1
    }
}
```

After this, one can simply inject it into the library by modifying **RxBluetoothKitLogger.defaultLogger** variable. A good place for injecting it can be the `AppDelegate.swift`.

Issue #344, opened 2019-06-29
L2CAP method openL2CAPChannel should be available for OSX 10.14


c0diq:
**Describe the bug**
Rx Peripheral openL2CAPChannel method should be available on OSX 10.14 and above.

**Expected behavior**
The CoreBluetooth method is available on 10.14 but the RxBluetoothkit version is missing the proper @available annotation.

**Environment:**
 - OS: OSX 10.14
 - Library version 5.1.2
 - Swift version 5.0

minixT:
Added to version 5.3.0

PR #345, opened 2019-07-02
Fix L2CAP API to support OSX 10.14

c0diq:
Fix for issue #344


Source/Peripheral.swift
```diff
@@ -829,8 +829,7 @@ public class Peripheral {
     /// * `BluetoothError.bluetoothPoweredOff`
     /// * `BluetoothError.bluetoothInUnknownState`
     /// * `BluetoothError.bluetoothResetting`
-    #if os(iOS) || os(tvOS) || os(watchOS)
```
dariuszseweryn: Shouldn't it stay? Currently probably it is not needed — technically though it is possible that there will be a new system that will not support the function?

c0diq: It’s following the same pattern as https://github.com/c0diq/RxBluetoothKit/blob/d710c2e105982dc503cb8073c3ff227e95a8b6ae/Source/Peripheral.swift#L418

PR #368, opened 2020-03-26
Fixed an issue where the SPM package doesn't compile with Xcode 11.4

akmarinov:
After updating to Xcode 11.4, my project is showing the following issue with RxBluetoothKit:

```
xcodebuild: error: Could not resolve package dependencies:
   target at '/Users/xcode/Library/Developer/Xcode/DerivedData/project-btxgpemzyyretjcousmtwoebmywz/SourcePackages/checkouts/RxBluetoothKit' contains mixed language source files; feature not supported
```

I've been able to track down the error to the [Info.plist](https://github.com/Polidea/RxBluetoothKit/blob/master/Source/Info.plist) file in the Sources folder.

For whatever reason, Xcode 11.4 now doesn't like that there's a file that's not a `.swift` one. (Xcode 11.3.1 still works fine)

I've included that file in the excluded section in the package description, which solves the issue.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -21,7 +21,8 @@ let package = Package(
                 exclude: [
                     "Example",
                     "Tests",
-                    "Source/Info.plist"
+                    "Source/Info.plist",
+                    "Source/RxBluetoothKit.h"
                 ],
                 sources: [
                     "Source"
--- a/Package.swift
+++ b/Package.swift
@@ -20,7 +20,8 @@ let package = Package(
                 path: ".",
                 exclude: [
                     "Example",
-                    "Tests"
+                    "Tests",
+                    "Source/Info.plist"
                 ],
                 sources: [
                     "Source"


DevAndArtist:
@akmarinov are you sure it's only `Info.plist`? The `Sources` folder does also contain `RxBluetoothKit.h`. I'd say SPM would still consider this as mixed languages. We should exclude this file as well, just in case.

akmarinov:
Well, Xcode works in mysterious ways, as the ObjC compatibility header doesn't seem to be an issue in my case.

```
// swift-tools-version:5.1
// The swift-tools-version declares the minimum version of Swift required to build this package.

import PackageDescription

let package = Package(
    name: "TestProject",
    platforms: [
        .iOS("12.4")
    ],
    products: [
        // Products define the executables and libraries produced by a package, and make them visible to other packages.
        .library(
            name: "TestProject",
            targets: ["TestProject"]),
    ],
    dependencies: [
        .package(url: "https://github.com/akmarinov/RxBluetoothKit.git", from: "5.3.1"),
        .package(url: "https://github.com/ReactiveX/RxSwift.git", from: "5.1.0")
    ],
    targets: [
        // Targets are the basic building blocks of a package. A target can define a module or a test suite.
        // Targets can depend on other targets in this package, and on products in packages which this package depends on.
        .target(
            name: "TestProject",
            dependencies: ["RxBluetoothKit", "RxSwift"],
            path: "Sources")
    ]
)
```

These package dependencies in a test project work without Xcode showing the issue for me, with the `RxBluetoothKit.h` included.

I'm not sure what the impact of turning off the compatibility header would be for ObjC projects.

DevAndArtist:
> excluding it in the SPM package shouldn't be an issue
That's my whole point. :)

tomovicn:
@DevAndArtist can you approve this pull request? Or who should we ping?
This is a blocker for all of us who use SPM.

DevAndArtist:
@tomovicn I'm just a contributor to the framework (I added some changes to the logging stuff in the past), but not part of Polidea so I can't really approve the PR. Personally I consider this framework as dead and will try to get rid of its dependency in our projects.

tomovicn:
It's a shame since it is a very good library. I dropped them a message to see if they still support this project.
Thanks for your response @DevAndArtist

minixT (approved):

PR #379, opened 2020-08-28
Correct Min MacOS version for CoreBluetooth Framework

walkerben:
'CBCentralManagerRestoredStatePeripheralsKey' is only available in macOS 10.13 or newer

CBCentralManagerRestoredStatePeripheralsKey is used in CentralManagerRestoredState.swift. Currently, compiling for macOS causes errors because the min macOS version is set to 10.10 and CBCentralManagerRestoredStatePeripheralsKey is not available.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -6,7 +6,7 @@ import PackageDescription
 let package = Package(
     name: "RxBluetoothKit",
     platforms: [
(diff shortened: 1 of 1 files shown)


minixT (approved):
Thanks for finding this.

PR #384, opened 2020-10-02
Optional CoreBluetooth parameters in CentralManager and PeripheralManager inits.

### LINKED ISSUE #364: Add initializer to CentralManager with CBCentralManager
**Is your feature request related to a problem? Please describe.**
I develop SDK (iOS/Cocoatouch framework) for custom earbuds used by iOS apps. RxBluetoothKit is used under the hood.

The main problem is that there is no possibility of retrieving peripherals. What? Why?

The app may use CoreBluetooth for personal needs. This means it creates CBCentralManager, discovers devices, and sets peripherals/passes identifiers to SDK to perform some operations.

RxBluetoothKit creates its own CBCentralManager behind the scene.

The problem is that data & events of CoreBluetooth are not shared between multiple instances of CBCentralManager.

If SDK calls retrievePeripherals(by ids:), observeConnect() or whatever functions using RxBluetoothKit (with own CBCentralManager) there is no effect.

**Describe the solution you'd like**
I would like to add an additional initializer to CentralManager so the client can pass a CBCentralManager instance and eliminate a hidden dependency.

```
public init(centralManager: CBCentralManager) {
        let delegateWrapper = CBCentralManagerDelegateWrapper()
        let centralManager = CBCentralManager(delegate: delegateWrapper, queue: queue, options: options)
        self.init(
            centralManager: centralManager,
            delegateWrapper: delegateWrapper,
            peripheralProvider: PeripheralProvider(),
            connector: Connector(centralManager: centralManager, delegateWrapper: delegateWrapper)
        )
    }
```

**Describe alternatives you've considered**
I understand that RxBlueetoothKit does not expose CoreBluetooth to clients, that's great 👍

I just want to mention a negative impact of hiding CBCentralManager inside of RxBluetoothKit:
- It looks like it's a best practice to have only one instance of CBCentralManager in the app. This is not currently possible if you want to create a framework that should be used by other clients.
- SDK cannot force iOS apps (clients) to use RxBluetoothKit.

Since this functionality may not be needed for all clients of RxBluetoothKit, we can create an additional CocoaPods subspec/module/whatever that contains this initializer.

For instance, [Moya](https://github.com/Moya/Moya#moya-version-vs-swift-version) provides additional functionality such as Rx & Reactive versions.

monikamsc:
Added optional CBCentralManager parameter to CentralManager init.
Added optional CBPeripheralManager parameter to PeripheralManager init.

Closes #364


Source/CentralManager.swift
```diff
@@ -70,10 +70,14 @@ public class CentralManager: ManagerType {
...
+    /// - parameter cbCentralManager: Optional instance of `CBCentralManager` to be used as a `manager`. If you
+    /// skip this parameter, there will be created an instance of `CBCentralManager` using given queue and options.
     public convenience init(queue: DispatchQueue = .main,
-                            options: [String: AnyObject]? = nil) {
+                            options: [String: AnyObject]? = nil,
+                            cbCentralManager: CBCentralManager? = nil) {
         let delegateWrapper = CBCentralManagerDelegateWrapper()
-        let centralManager = CBCentralManager(delegate: delegateWrapper, queue: queue, options: options)
+        let centralManager = cbCentralManager != nil ?
+            cbCentralManager! : CBCentralManager(delegate: delegateWrapper, queue: queue, options: options)
```
minixT, line 80:
```suggestion
        let centralManager = cbCentralManager ?? CBCentralManager(delegate: delegateWrapper, queue: queue, options: options)
```
This will allow you not to use force unwrapping.

minixT (changes requested):

minixT (approved):

Issue #394, opened 2021-03-10
Support building xcframeworks via carthage


badeleux:
**Describe the bug**
Can't compile the library into xcframework due to following issue:

```
import RxSwift
       ^

** ARCHIVE FAILED **


The following build commands failed:
	CompileSwift normal arm64
	CompileSwiftSources normal arm64 com.apple.xcode.tools.swift.compiler
	CompileSwift normal armv7
	CompileSwiftSources normal armv7 com.apple.xcode.tools.swift.compiler
(4 failures)
```

**To Reproduce**
1. Try to run a carthage on a project which uses RxBluetoothKit
2. `carthage bootstrap --no-use-binaries --platform iOS --use-xcframeworks`

**Expected behavior**

**Environment:**
 - Device: Macbook with M1
 - OS: MacOS 11.2
 - Library version 6.0.0
 - Swift version 5.3

PR #395, opened 2021-03-10
#394 Support xcframeworks build

badeleux:
https://github.com/Polidea/RxBluetoothKit/issues/394


David2XN:
Using XCFrameworks is now highly encouraged by Apple/Carthage. This is a must-have feature.

